PR #5, opened 2021-06-23
added possibility to work with scalebars and scaling factors

linuxrider:
The plots in `sato_effect_2006_725_p728_4a_red` (in our Google Drive) use scalebars/scaling factors that the digitizer did not understand yet.

Here we read the scale bar from a Bezier curve and the corresponding scaling factor from its title. (This is maybe not the right way to do this yet. Should we use xlink?)

We also start to use regular expressions to parse the different bits of text in the SVG.


saraedum:
Could you add the SVG (without the PNG) here as well? So we can maybe turn it into an automated test?

svgdigitizer.py
```diff
@@ -18,8 +24,13 @@ def __init__(self, filename, xlabel=None, ylabel=None):
         
         self.ref_points, self.real_points = self.get_points()
         
-        self.xtrafo = self.get_trafo([self.ref_points['x1'][0], self.ref_points['x2'][0]], [self.real_points['x1'], self.real_points['x2']])
-        self.ytrafo = self.get_trafo([self.ref_points['y1'][1], self.ref_points['y2'][1]], [self.real_points['y1'], self.real_points['y2']])
+        self.scalebars = self.get_scalebars()
+        self.scaling_factors = self.get_scaling_factors()
```
saraedum, line 28: Maybe it's easier to turn these two into properties instead? I.e., write

```
@cached_property
def scalebars(self):
    ...
```

svgdigitizer.py
```diff
@@ -32,38 +43,83 @@ def get_points(self):
...
+
     def get_parsed(self):
         '''cuve function'''
         self.allresults = {}
         for pathid, pvals in self.figpaths.items():
             self.allresults[pathid] = self.get_real_values(pvals)
     
 
-    def get_trafo(self, xrefpath, xrefvalues):
+    def get_trafo(self, axis):
```
saraedum, line 105: What kind of transformations can happen here? I assume it could be any [linear affine transformation](https://en.wikipedia.org/wiki/Affine_transformation). Should we then maybe use a matrix to represent it?

saraedum (commented):
I had a look at the diagrams from the Sato paper but it is still not entirely clear to me what we are trying to achieve here. So, these are diagrams which show the scale of one axis somewhere in the picture and not along the axis. That's a "scalebar" I guess. But what is the scaling factor good for?

svgdigitizer.py
saraedum: We might consider doing this for most of the methods actually, so we get rid of state in the digitizer.

saraedum:
@linuxrider I took the liberty to rewrite your PR summary as I found it a bit hard to follow initially. I hope I did not break what you were trying to say.

svgdigitizer.py
linuxrider: I don't know the mathematical term for the transformation but of course I'm in favor of a mathematical sound solution. Is this solution also suitable for waterfall(/3D) graphs? (will lookup an example)

saraedum: I think so. These transformations contain shears and I think that's what you need for waterfalls.

svgdigitizer.py
```diff
@@ -18,8 +24,13 @@ def __init__(self, filename, xlabel=None, ylabel=None):
         
         self.ref_points, self.real_points = self.get_points()
         
-        self.xtrafo = self.get_trafo([self.ref_points['x1'][0], self.ref_points['x2'][0]], [self.real_points['x1'], self.real_points['x2']])
-        self.ytrafo = self.get_trafo([self.ref_points['y1'][1], self.ref_points['y2'][1]], [self.real_points['y1'], self.real_points['y2']])
+        self.scalebars = self.get_scalebars()
```
linuxrider, line 27:
```suggestion

```

svgdigitizer.py
linuxrider, line 28:
```suggestion

```

linuxrider: I agree.

linuxrider, line 28:
```suggestion

```

svgdigitizer.py
```diff
@@ -32,38 +43,83 @@ def get_points(self):
...
+                ref_point_id = regex_match.group("point")
+                real_points[ref_point_id] = float(regex_match.group("value"))
+                ref_points[ref_point_id] = {'x': float(text.firstChild.getAttribute('x')), 'y': float(text.firstChild.getAttribute('y'))}
+
         
         print('Ref points: ', ref_points)
         print('point values: ',real_points)
         return ref_points, real_points
         
+    def get_scalebars(self):
```
linuxrider, line 64:
```suggestion
    @cached_property
    def scalebars(self):
```

svgdigitizer.py
```diff
@@ -32,38 +43,83 @@ def get_points(self):
...
+                scalebars[regex_match.group("axis")] = {}
+                if regex_match.group("axis") == 'x':
+                    scalebars[regex_match.group("axis")]['ref'] = abs(parsed_path.point(1).real - parsed_path.point(0).real)
+                elif regex_match.group("axis") == 'y':
+                    scalebars[regex_match.group("axis")]['ref'] = abs(parsed_path.point(1).imag - parsed_path.point(0).imag)
+                scalebars[regex_match.group("axis")]['real'] = float(regex_match.group("value"))
+
+        return scalebars
+
+    def get_scaling_factors(self):
```
linuxrider, line 85:
```suggestion
    @cached_property
    def scaling_factors(self):
```

svgdigitizer.py
```diff
@@ -4,6 +4,12 @@
 import numpy as np
 import pandas as pd
```
linuxrider, line 6:
```suggestion
from cached_property import cached_property

```

svgdigitizer.py
```diff
@@ -32,38 +43,83 @@ def get_points(self):
...
-        trafo = lambda xpathdata: mrefx * (xpathdata - xrefpath[0]) + xrefvalues[0]
+        # value = mref * (path - refpath0 ) + refvalue0
+        #print('refpath ', refpath, 'refvalues ', refvalues)
+        p_real = self.real_points
+        p_ref = self.ref_points
+        try:
+            mref = (p_real[f'{axis}2'] - p_real[f'{axis}1']) / (p_ref[f'{axis}2'][axis] - p_ref[f'{axis}1'][axis]) / self.scaling_factors[axis]
+            trafo = lambda pathdata: mref * (pathdata - p_ref[f'{axis}1'][axis]) + p_real[f'{axis}1']
+        except KeyError:
+            print(self.scaling_factors[axis])
```
linuxrider, line 115:
```suggestion

```

svgdigitizer.py
```diff
@@ -32,38 +43,83 @@ def get_points(self):
...
+    def get_scaling_factors(self):
+        scaling_factors = {'x': 1, 'y': 1}
+        for text in self.doc.getElementsByTagName('text'):
+
+            # parse text content
+            text_content = text.firstChild.firstChild.data
+            regex_match = re.match(scaling_factor_regex_str, text_content)
+
+            if regex_match:
+                print(regex_match.group("value"))
```
linuxrider, line 94:
```suggestion

```

tests/test_regex.py
```diff
@@ -0,0 +1,58 @@
+import unittest
+import re
+from digitizer import ref_point_regex_str, scalebar_regex_str, scaling_factor_regex_str
```
linuxrider, line 3: Maybe the import statement needs to be changed

saraedum:
@linuxrider, I think you need to merge in the origin/main branch here and resolve conflicts. Let me know if you need help with this.

linuxrider:
@saraedum I tried to follow the "command line instructions" but it did not resolve the conflict.
I would need your help to accomplish resolving the conflict.

linuxrider:
With the help of @saraedum I was able to resolve the merge conflict.
Finally test are passing again.

PR #9, opened 2021-07-01
add helper for pdf2png/svg conversion

linuxrider:
The idea is to simply run `preparesvg` on the pdf file to end up with ready to use svgs.
~~Although `python -m svgdigitizer.helpers.converter tosvg <pdf file>` runs smoothly `preparesvg tosvg` has an import error.
It is unclear to me why because I follow the entrypoint scheme of @saraedum.~~


linuxrider:
Never mind, `preparesvg` works, `pip install -e .` did the trick.

saraedum:
I think I would prefer to put everything into the svgdigitizer CLI if possible so we do not clutter the path with commands. I think we could have nested commands such as `svgdigitizer prepare topng ...` but maybe just `svgdigitizer paginate --png input.pdf` is good enough for the time being?

setup.py
```diff
@@ -14,10 +14,14 @@
         'pandas',
         'click',
         'pyyaml',
+        'pillow',
```
saraedum, line 17: Note that you also need to add these to `environment.yml` so they get installed in the automated tests.

PR #20, opened 2021-07-10
rudimentary x sampling by add --sampling_interval parameter

linuxrider:
This should enable svgdigitizer to ignore the points of the bezier curves and sample it by the --sampling_interval parameter in mV.


svgdigitizer/__main__.py
```diff
@@ -4,16 +4,24 @@
 def cli(): pass
 
 @click.command()
+@click.option('--sampling_interval',  help='specify sampling interval (for now in mV)')
```
saraedum, line 7: I think you should add a `sampling_interval` parameter to the `def` below. You can also specify here that this parameter is a `float` with the default `None`. Then you don't need any of the `if/else` logic below.

svgdigitizer/__main__.py
```diff
@@ -4,16 +4,24 @@
...
 def plot(svg):
     from svgdigitizer.svgdigitizer import SvgData
-    SvgData(svg).plot()
+    if sampling_interval:
+        SvgData(svg, sampling_interval=float(sampling_interval)).plot()
+    else:
+        SvgData(svg).plot()
 
 @click.command()
+@click.option('--sampling_interval',  help='specify sampling interval (for now in mV)')
```
saraedum, line 17: Same comment applies to the `sampling_interval` here.

svgdigitizer/svgdigitizer.py
```diff
@@ -12,29 +13,34 @@
...
-    def __init__(self, filename, xlabel=None, ylabel=None):
+    def __init__(self, filename, xlabel=None, ylabel=None, sampling_interval=None):
         '''filename: should be a valid svg file created according to the documentation'''
         self.filename = filename
-        
+
         self.xlabel = xlabel or 'x'
-        self.ylabel = ylabel or 'y'
-        
+        self.ylabel = ylabel or 'y'
```
saraedum, line 21:
```suggestion
        self.ylabel = ylabel or 'y'
```

svgdigitizer/svgdigitizer.py
```diff
@@ -164,9 +170,12 @@ def get_paths(self):
         for path in paths:
             if path.parentNode == layer:
                 path_strings.append((path.getAttribute('id'), path.getAttribute('d')))
-
-        xypaths_all = {path_string[0]: self.parse_pathstring(path_string[1]) for path_string in path_strings}
-
+        if not self.sampling_interval:
+            print('parsing curve')
```
saraedum, line 174:
```suggestion
```

svgdigitizer/svgdigitizer.py
```diff
@@ -164,9 +170,12 @@ def get_paths(self):
...
+            xypaths_all = {path_string[0]: self.parse_pathstring(path_string[1]) for path_string in path_strings}
+        elif self.sampling_interval:
+            print('sampling curve')
```
saraedum, line 177:
```suggestion
```

svgdigitizer/svgdigitizer.py
```diff
@@ -181,7 +190,40 @@ def parse_pathstring(self, path_string):
...
+        xmin, xmax, ymin, ymax = path.bbox()
+
+        points = []
+        for segment in path:    
+            segment_points = [[],[]]    
+            for x in np.linspace(xmin, xmax, int(abs(xmin - xmax)/self.transformed_sampling_interval)):
+                intersects = Path(Line(complex(x,ymin),complex(x,ymax))).intersect(Path(segment))
+                # it is possible that a segment includes both scan directions
+                # which leads to two intersections
+                #print(intersects)
```
saraedum, line 207:
```suggestion
```

svgdigitizer/__main__.py
linuxrider: Yes, is more elegant. Thanks.

linuxrider:
It is unclear to me why the sampling test fails since it is working on my machine and the yaml file is the same as for scalebar test.

linuxrider:
Sampling test fails due to KeyError in sampling-test although the same yaml works in scalebar-test.

saraedum:
@linuxrider could you resolve the merge conflicts here?

linuxrider:
Although some tests are failing (due to path detection logic) I will merge this since we are change the detection logic anyway.

svgdigitizer/__main__.py
```diff
@@ -4,16 +4,18 @@
...
+@click.option('--sampling_interval', type=float, default=None, help='specify sampling interval (for now in mV)')
 @click.argument('svg')
-def plot(svg):
+def plot(svg, sampling_interval):
     from svgdigitizer.svgdigitizer import SvgData
-    SvgData(svg).plot()
+    SvgData(svg, sampling_interval=sampling_interval).plot()
 
 @click.command()
+@click.option('--sampling_interval', type=float, default=None, help='specify sampling interval (for now in mV)')
```
saraedum, line 14: It's weird that we now have to specify all the parameters more than once.

Maybe, we should use global parameters to configure `SvgData`? I.e., maybe `CreateCVData` should be refactored so it is fed a `SvgData` object instead of taking any parameters itself?

svgdigitizer/svgdigitizer.py
```diff
@@ -12,25 +13,31 @@
...
             self.trafo[axis] = self.get_trafo(axis)
 
+        self.sampling_interval = sampling_interval
+
         self.get_parsed()
         self.create_df()
     
+    @cached_property
+    def transformed_sampling_interval(self):
+        factor = 1/((self.trafo['x'](self.sampling_interval)-self.trafo['x'](0))/(self.sampling_interval/1000))
```
saraedum, line 38: I think you can just write `return` here without using a temporary.

svgdigitizer/svgdigitizer.py
```diff
@@ -183,7 +191,44 @@ def parse_pathstring(self, path_string):
...
+                        segment_points[i].append((point.real, point.imag))   
+
+            # second intersection is appended in reverse order!! 
+            if len(segment_points[1]) > 0:    
+                segment_points[0].extend(segment_points[1][::-1]) 
+            # sometimes segments are shorter than sampling interval   
+            if len(segment_points[0]) > 0:    
+                first_segment_point = (segment.point(0).real, segment.point(0).imag)    
+
+                if (((segment_points[0][-1][0]-first_segment_point[0])**2+(segment_points[0][-1][1]-first_segment_point[1])**2)**0.5 >
```
saraedum, line 224: I think there is no need for a square root here.

saraedum (commented):
Could we fix this up somehow so the tests pass again in particular?

svgdigitizer/__main__.py
```diff
@@ -4,16 +4,18 @@
 def cli(): pass
 
 @click.command()
+@click.option('--sampling_interval', type=float, default=None, help='specify sampling interval (for now in mV)')
```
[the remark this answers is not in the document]
linuxrider: @DunklesArchipel How should we handle this to stay generic?
We could explicitly specify the unit?

svgdigitizer/svgdigitizer.py
linuxrider: I'm not sure if there are conditions where we need Pythagoras.

PR #22, opened 2021-07-11
change path detection logic

saraedum:
since we cannot reliably detect inkscape layers, we now treat anything
containing more than one path as a layer.


linuxrider:
Maybe, we should just make it more explicit.
We could group the measruement path with an identifier text. Something like
"curve: <identifier as in the yaml>"
In principal, all paths then have an identifier.

PR #25, opened 2021-07-14
change to text label based path detection

linuxrider:


svgdigitizer/svgdigitizer.py
```diff
@@ -10,6 +10,7 @@
 ref_point_regex_str = r'^(?P<point>(x|y)\d)\: ?(?P<value>-?\d+\.?\d*) ?(?P<unit>.+)?'
 scale_bar_regex_str = r'^(?P<axis>x|y)(_scale_bar|sb)\: ?(?P<value>-?\d+\.?\d*) ?(?P<unit>.+)?'
 scaling_factor_regex_str = r'^(?P<axis>x|y)(_scaling_factor|sf)\: (?P<value>-?\d+\.?\d*)'
+curve_regex_str = r'^curve: ?(?P<curve_id>[\W|\w]*)'
```
saraedum, line 13: What is `\w|\W`? Is this not the same as `.`?

svgdigitizer/svgdigitizer.py
```diff
@@ -157,20 +158,20 @@ def paths(self):
...
-            if path.parentNode.nodeName != 'g' or
-                # This is complicated by the fact that layers as created by inkscape
-                # are also groups. Such layers have the 'inkscape:groupmode' set.
-                path.parentNode.getAttribute("inkscape:groupmode") == 'layer' or
-                # But this attribute goes away when exporting to plain SVG. In
-                # such case, we recover layers as the groups that contain more
-                # than one path.
-                len(path.parentNode.getElementsByTagName("path")) > 1
-        }
+        paths = {}
```
saraedum, line 161: Have we now decided that there can be no `<path>` unless it is grouped with some text? If so, we could probably refactor things a bit here by introducing a method that returns all the pairs `(text string, path)` for all the paths in the SVG (and warns about paths that are not following this scheme.)
This function could then be used everywhere and simplify the code that searches the groupings.

svgdigitizer/svgdigitizer.py
linuxrider: I think you are right. Changed.

svgdigitizer/svgdigitizer.py
linuxrider: Yes, I think we did. @DunklesArchipel

svgdigitizer/svgdigitizer.py
```diff
@@ -157,29 +149,48 @@ def get_real_values(self, xpathdata):
...
+    @cached_property
+    def labeled_paths(self):
+        r"""
+        Return all paths which are grouped/label with text elements, warn if
+        unlabeled or ambiguous paths exist.
+        """
+        labeled_paths = {}
+        texts = self.doc.getElementsByTagName('text')
+        for text in texts:
+
```
saraedum, line 161:
```suggestion

        for text in texts:
```

svgdigitizer/svgdigitizer.py
```diff
@@ -157,29 +149,48 @@ def get_real_values(self, xpathdata):
...
+            group = text.parentNode
+            paths = group.getElementsByTagName('path')
+            # warn for ambiguous paths
+            # gives false positive for scale bars
+            #if len(paths) > 1:
+            #    print("""Multiple paths grouped with single text!
+#Only first will be taken into account. Please review supplied svg file!""")
```
saraedum, line 168: We should do something about this so this also works with scalebars.
Can we include the actual text in the message? I think that makes debugging much easier.

svgdigitizer/svgdigitizer.py
```diff
@@ -157,29 +149,48 @@ def get_real_values(self, xpathdata):
...
+    def labeled_paths(self):
+        r"""
+        Return all paths which are grouped/label with text elements, warn if
+        unlabeled or ambiguous paths exist.
+        """
+        labeled_paths = {}
+        texts = self.doc.getElementsByTagName('text')
+        for text in texts:
+
+            group = text.parentNode
```
saraedum, line 162: Can we check that this is actually a group?

svgdigitizer/svgdigitizer.py
```diff
@@ -11,6 +11,7 @@
 ref_point_regex_str = r'^(?P<point>(x|y)\d)\: ?(?P<value>-?\d+\.?\d*) ?(?P<unit>.+)?'
 scale_bar_regex_str = r'^(?P<axis>x|y)(_scale_bar|sb)\: ?(?P<value>-?\d+\.?\d*) ?(?P<unit>.+)?'
 scaling_factor_regex_str = r'^(?P<axis>x|y)(_scaling_factor|sf)\: (?P<value>-?\d+\.?\d*)'
+curve_regex_str = r'^curve: ?(?P<curve_id>.*)'
```
saraedum, line 14: We could consider putting all these regexes into a dict, say `label_patterns`. Then we could further down also check that all labels of paths match exactly one of these regular expressions and warn otherwise.

saraedum: In fact, `labeled_paths` could then already return a dict of the form
```
{
  "ref_point": [(text,path), (text,path)]
}
```
and all the regex handling would be confined to `labeled_paths` and we do not need to repeat it in several places.
```

File: svgdigitizer/svgdigitizer.py (hunk 5 above)
Comment (linuxrider) [reply svgdigitizer/svgdigitizer.py]: I think the parent of a text is always a group since layers are also groups.

File: svgdigitizer/svgdigitizer.py (hunk 6 above)
Comment (linuxrider) [reply svgdigitizer/svgdigitizer.py]: Have you thought of something like in `try with label_pattern dict`?

File: svgdigitizer/svgdigitizer.py (hunk 4 above)
Comment (linuxrider) [reply svgdigitizer/svgdigitizer.py]: One "easy" way would be to switch back to a single path per scale bar.

File: svgdigitizer/svgdigitizer.py (hunk 5 above)
Comment (saraedum) [reply svgdigitizer/svgdigitizer.py]: I think this is not true for top-level texts.

File: svgdigitizer/svgdigitizer.py (hunk 6 above)
Comment (saraedum) [reply svgdigitizer/svgdigitizer.py]: I don't understand the question, sorry.

Comment (linuxrider) [reply svgdigitizer/svgdigitizer.py]: In the commit `try with label_pattern dict` I tried to implement the idea you mentioned above.
Now the question is if you think the implementation fits your idea?

File: svgdigitizer/svgdigitizer.py (hunk 5 above)
Comment (linuxrider) [reply svgdigitizer/svgdigitizer.py]: It works without a layer. In this case, top_level is 'svg'.
Top-level texts are excluded in a later commit.

File: svgdigitizer/svgdigitizer.py [hunk 7]
```diff
@@ -157,29 +148,87 @@ def get_real_values(self, xpathdata):
...
+        """
+
+        labeled_paths = {}
+
+        # determine top level group (layer or svg) by counting groups
+        # group with most subgroups is probably the relevant top level
+        svg = self.doc.getElementsByTagName('svg')[0]
+        group_counts = {}
+
+        for group in svg.getElementsByTagName('g'):
```
linuxrider, line 165:
```suggestion

```

PR #36, opened 2021-07-26
Do not use filenames explicitly

### LINKED ISSUE #34: Drop create_csv function
I think we should actually just drop such functions and just document how to do this instead.

_Originally posted by @saraedum in https://github.com/echemdb/svgdigitizer/pull/33#discussion_r675465586_

saraedum:
Instead, create `SVGPlot` from an arbitrary stream. I added some trivial doctests for this (and rudimentary documentation for `SVGPlot`.) The doctests initially failed, so I had to make some parts of the code (`labeled_paths`) more flexible.

Also fixes #34, #37.


svgdigitizer/svgplot.py
```diff
@@ -16,18 +15,63 @@
...
 
+logger = logging.getLogger('svgplot')
+
 class SVGPlot:
-    def __init__(self, filename, xlabel=None, ylabel=None, sampling_interval=None):
+    r"""
+    A plot as a Scalable Vector Graphics (SVG) which can be converted to (x, y)
+    coordinate pairs.
+
+    Typically, the SVG input has been created a tracing a measurement plot from
```
linuxrider, line 25:
```suggestion
    Typically, the SVG input has been created by tracing a measurement plot from
```

svgdigitizer/svgplot.py
```diff
@@ -143,67 +184,92 @@ def get_trafo(self, axis):
...
+
     def get_real_values(self, xpathdata):
         xnorm = self.trafo['x'](xpathdata[:, 0])
         ynorm = self.trafo['y'](xpathdata[:, 1])
         return np.array([xnorm, ynorm])
 
+    @classmethod
+    def _text_value(cls, node):
+        r"""
+        Return the text content of a node such as a `<text>` node.
```
linuxrider, line 196:
```suggestion
        Return the text content of a node (including the text of its children) such as a `<text>` node.
```

svgdigitizer/svgplot.py
```diff
@@ -143,67 +184,92 @@ def get_trafo(self, axis):
...
+                    if SVGPlot._text_value(child):
+                        logger.warning(f'Ignoring unexpected text node "{SVGPlot._text_value(child)}" grouped with <path>.')
+                elif child.nodeType == Node.ELEMENT_NODE:
+                    if child.tagName == 'path':
+                        continue
+                    if child.tagName != "text":
+                        logger.warning(f"Unexpected <{child.tagName}> grouped with <path>. Ignoring unexpected <{child.tagName}>.")
+                        continue
+
+                    if label is not None:
```
linuxrider, line 255: `label` is always `None`

saraedum: No, it's set below where it says `label = child`. So this will complain when there are multiple `<text>` nodes in a group.

svgdigitizer/svgplot.py
```diff
@@ -36,15 +80,15 @@ def __init__(self, filename, xlabel=None, ylabel=None, sampling_interval=None):
...
     @cached_property
     def transformed_sampling_interval(self):
         factor = 1/((self.trafo['x'](self.sampling_interval)-self.trafo['x'](0))/(self.sampling_interval/1000))
         return factor
-    
+
     def get_points(self):
         '''Creates:
-        ref_points: relative values of the spheres/eclipses in the svg file.
+        ref_points: relative values of the spheres/ellipses in the svg file.
```
DunklesArchipel, line 91: Actually which spheres/ellipses. I think this is a relic from the first modified version, where spheres were located on the axis.

linuxrider (approved):

PR #49, opened 2021-08-03
Use matrices to represent map from SVG to plot coordinate system

### LINKED ISSUE #40: SVGPlot cannot handle multiple curves properly
Currently, `SVGPlot` creates a list of dataframes, one for each curve. However, these dataframes are not labeled so there is no way to recover which data frame belong to which `curve: ...` identifier.

We should probably change this so that `SVGPlot.__init__` gets another parameter `curve=None` which selects which single curve is parsed. If the value is `None`, it selects the first curve and requires it to be unique.
If somebody wants to handle multiple curves in the same SVG, they'd have to create multiple `SVGPlot` instances, one for each curve. However, we are not planning to use this feature ourselves.

### LINKED ISSUE #50: SVGPlot.parse_path() drops the last point
Currently, `SVGPlot.parse_path()` reports the first point twice and drops the last point.

E.g., when calling it with `M 0 0 L 1 1` it returns `[[0, 0], [0, 0]]`.

saraedum:
This approach is more flexible as it can also handle rotated plots (e.g., when a plot was photographed or manually scanned) and plots with non-orthogonal axes (probably not very common but maybe plot axes are not exactly orthogonal in a scan.)

The matrix approach is flexible enough so that the old algorithm can be recovered as a special case. Anyway, with the new algorithm (the one that supports rotated pictures,) some of the tests change their output.

In the case `xy.svg`, the labels on the axes are:

| Label | Value | SVG x Coordinate | SVG y Coordinate |
| --- | --- | --- | --- |
| x1 | 0.0 | 24.182038 | 77.082165 |
| x2 | 0.6 | 55.754552 | 76.956168 |
| y1 |-0.2 | 67.358368 | 145.88578 |
| y2 | 0.1 | 67.501698 | 126.05364 |

Let's apply the deduced transformation to some points:

| SVG Coordinates | Transformed Coordinate (old) | Transformed Coordinate (new) |
| --- | --- | --- |
| (24.182038, 145.88578) | (0.0, -0.2) | (0.0095, -0.1973) |
| (55.754552, 126.05364) | (0.6, 0.1) | (0.6067, 0.1007) |
| (24.182038, 77.082165) | (0, 0.8408) | (0, 0.8434) |
| (55.754552, 76.956168) | (0.6, 0.8427) | (0.6, 0.8434) |

What this means: The old algorithm assumes that the plot is perfectly axis-aligned with the SVG coordinate system. The new algorithm assumes that the end points of the markers are exactly on the (same) axis. When this is not exactly the case, the algorithm detects that the plot is slightly rotated.

Which one is better? This is a matter of taste. Probably the old algorithm is less surprising, so it will be the default at the moment. However, it does not work well when the picture is somewhat rotated or the axes are not exactly orthogonal.

also fixes #40.
also fixes #50.


saraedum:
With these changes, some of the tests change their output.
In the case `xy.svg`, the labels on the axes are:


| Label | Value | SVG x Coordinate | SVG y Coordinate |
| --- | --- | --- | --- |
| x1 | 0.0 | 24.182038 | 77.082165 |
| x2 | 0.6 | 55.754552 | 76.956168 |
| y1 |-0.2 | 67.358368 | 145.88578 |
| y2 | 0.1 | 67.501698 | 126.05364 |

Let's apply the deduced transformation to some points:

| SVG Coordinates | Transformed Coordinate (before) | Transformed Coordinate (after) |
| --- | --- | --- |
| (24.182038, 145.88578) | (0.0, -0.2) | (0.0095, -0.1973) |
| (55.754552, 126.05364) | (0.6, 0.1) | (0.6067, 0.1007) |

.github/workflows/test.yml
```diff
@@ -31,27 +31,27 @@ jobs:
         shell: bash -l {0}
         run: |
           svgdigitizer digitize test/data/xy.svg
-          diff test/data/xy.csv test/data/xy.csv.expected
+          python -c 'import pandas; pandas.testing.assert_frame_equal(pandas.read_csv("test/data/xy.csv"), pandas.read_csv("test/data/xy.csv.expected"))'
```
saraedum, line 34: The old `diff` approach does not work anymore since the exact numeric output now apparently depends on the numpy version. Anyway, the differences are only in some digits that certainly have no meaning anymore as you can see in the changes to the CSV files below.
I had to update the CSVs because the first line was always duplicated and the last line was missing, see #50. I also updated the rest to the output I now get with Python 3.9. I checked that nothing substantial has changed there except for a few trailing digits.

saraedum:
It would be great if somebody could have a look at the changes here :)

svgdigitizer/svgplot.py
```diff
@@ -28,160 +28,514 @@ class SVGPlot:
...
+        ...     <path d="M 0 0 L 100 0" />
+        ...     <text x="0" y="0">y2: 1</text>
+        ...   </g>
+        ... </svg>''')
+        >>> SVGPlot(svg, algorithm='mark-aligned').transformation
+        array([[ 0.01,  0.01, -1.  ],
+               [ 0.  , -0.01,  1.  ],
+               [ 0.  ,  0.  ,  1.  ]])
+
+        """
```
saraedum, line 466: The algorithm below is basic linear algebra. It might still be unclear how it works if you don't do these things regularly. I tried to put a lot of comments in here. If you are interested how this actually works, I am happy to talk about it next time.

linuxrider (approved):

PR #53, opened 2021-08-06
Apply transformations to determine locations of points

### LINKED ISSUE #52: getAttribute cannot reliably determine coordinates
Currently, we often get coordinates of things such as text or paths with `getAttribute`, e.g., `getAttribute("x")` or `getAttribute("d")`. However, this produces the raw attribute value without any transformation applied. If, for example, we move a group by applying a transformation to the entire `<g>`, a contained `<path>` does not know about this and we parse its `d` as if it were in absolute SVG coordinates.

saraedum:
fixes #52, #54.


DunklesArchipel (commented):
looks good with my limited knowledge. :)

DunklesArchipel:
I guess you can merge this. It seems to do what it should.

linuxrider (approved):
Looks good to me.

PR #57, opened 2021-08-18
Implement faster sampling

saraedum:
We now get about 20k samples per second.

Here is a plot before/after or after/before, I forget. Anyway, nothing has changed it seems:

image

Note that this changes the API of SVGPlot somewhat. Now `.curve()` returns an SVG path in the plot coordinate system and not just a list of points anymore. The actual sampling only happens when you invoke `.df()` in some way.

Now, all the time is spent in finding roots of degree three polynomials. It should not be too hard to speed this up even further but I guess this is quite sufficient already.


saraedum:
Unfortunately, this is not as fast & stable as I would like it to be yet. There are no optimizations for 1-dimensional Bezier curves in svgpathtools so we need to roll a better version of `length` and `ilength` for this to really make sense.

saraedum:
The case of line segments and quadratic Bézier curves are probably very easy. In the cubic case we get to solve with the usual `B(t)=(1-t)^3P_0+3(1-t)^2tP_1+3(1-t)t^2P_2+t^3P_3` for `\int \pi_xB(t)dt = x`. The integral is just integrating a degree 3 polynomial so the integral is a degree 4 polynomial. Solving for a specific `x` can be done with Newton or any other fast approximation scheme.

There is also the possibilty of an Arc in SVG which is just a conic section I believe. It's probably irrelevant for our purposes. If we really need these it would not be too hard to implement I guess.

.github/workflows/test.yml
```diff
@@ -40,12 +40,12 @@ jobs:
       - name: sampling-test
         shell: bash -l {0}
         run: |
-          svgdigitizer digitize test/data/sampling.svg --sampling_interval 1
+          svgdigitizer digitize test/data/sampling.svg --sampling_interval .00101
```
saraedum, line 43: These constants were chosen such that the output CSV has the same number of rows than before.

environment.yml
```diff
@@ -17,4 +17,5 @@ dependencies:
   - svgpathtools
   - pycodestyle
   - pytest
+  - scipy
```
saraedum, line 20: some operations in svgpathtools are faster when this module is available.

svgdigitizer/__main__.py
```diff
@@ -21,7 +21,7 @@
 # ********************************************************************
 import click
 
-help_sampling = 'sampling interval (in mV)'
+help_sampling = 'sampling interval on the x-axis'
```
saraedum, line 24: The unit is now in the unit that is specified in the plot. So if the axis is in mV nothing changes.

svgdigitizer/svgplot.py
```diff
@@ -738,7 +718,7 @@ def curve(self):
         ... </svg>'''))
         >>> plot = SVGPlot(svg)
         >>> plot.curve
-        [(0.0, 100.0), (100.0, 0.0)]
+        Path(Line(start=0j, end=(1+1j)))
```
saraedum, line 721: This now returns the actual curve in the plot coordinate system, i.e., here the line going from (0, 0) to (1, 1).

svgdigitizer/svgplot.py
```diff
@@ -798,47 +775,303 @@ def labeled_paths(self):
...
+
+        # The current path length on the x-axis at which we plan to sample (in the range [0, length of the current path segment]):
+        X = 0
+
+        # A projection down to the x-axis so that we can measure lengths in
+        # x-axis direction only.
+        from numpy import identity
+        project_x = identity(3)
+        project_x[1][1] = 0
+
```
saraedum, line 870:
```suggestion
```
Not used anymore.

PR #62, opened 2021-08-27
Update readme for CLI 

DunklesArchipel:
I updated the CLI readme to include the CV Option.

Are the examples provided in the readme useful?
For example `digitize` and `cv` show the same results.

Furthermore, should the options also be shown in the examples?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ Options:
   --help  Show this message and exit.
 
 Commands:
+  cv
   digitize
   paginate
   plot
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,6 +34,8 @@ $ svgdigitizer digitize test/data/Ni111_NaOH_Beden1985_Fig2c.svg
 [creates test/data/Ni111_NaOH_Beden1985_Fig2c.csv]
 $ svgdigitizer paginate <pdf file>
 [paginates the pdf file]
+$ svgdigitizer cv test/data/Ni111_NaOH_Beden1985_Fig2c.svg
+[creates test/data/Ni111_NaOH_Beden1985_Fig2c.csv]
```
saraedum, line 38: We could distinguish the two `.csv` outputs (this one and the one of `digitize`) by saying something like `creates ...csv with axes x,y,z`?

saraedum (approved):

saraedum:
I think these examples are useful and I would essentially keep them the way they are.

PR #64, opened 2021-09-12
extend refpoint regex for references

linuxrider:


linuxrider:
The reference is now passed down to cv just like a unit.
Is this a good way?
Had to change doctests due to a wired fail:
```__________________________________________________________________________________ [doctest] svgdigitizer.svgplot.SVGPlot.units ___________________________________________________________________________________
EXAMPLE LOCATION UNKNOWN, not showing all tests of that example
??? >>> with TestCase.assertLogs(_) as logs:
??? ...    plot.units
??? ...    print(logs.output)
UNEXPECTED EXCEPTION: AttributeError("'dict' object has no attribute '_formatMessage'")
  File "/home/hermann/nextcloud/coding/envs/miniconda/envs/python/lib/python3.9/doctest.py", line 1336, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest svgdigitizer.svgplot.SVGPlot.units[15]>", line 3, in <module>
  File "/home/hermann/nextcloud/coding/envs/miniconda/envs/python/lib/python3.9/unittest/_log.py", line 67, in __exit__
    self._raiseFailure(
  File "/home/hermann/nextcloud/coding/envs/miniconda/envs/python/lib/python3.9/unittest/case.py", line 162, in _raiseFailure
    msg = self.test_case._formatMessage(self.msg, standardMsg)
AttributeError: 'dict' object has no attribute '_formatMessage'
/home/hermann/nextcloud/coding/svgdigitizer/svgdigitizer/svgplot.py:None: UnexpectedException
```
Maybe it is related to the handling of scale bar labels?

svgdigitizer/svgplot.py
```diff
@@ -785,9 +864,9 @@ def labeled_paths(self):
         are going to extract the (x, y) coordinate pairs from.
         """
         patterns = {
-            'ref_point': r'^(?P<point>(x|y)\d)\: ?(?P<value>-?\d+\.?\d*) *(?P<unit>.+)?',
-            'scale_bar': r'^(?P<axis>x|y)(_scale_bar|sb)\: ?(?P<value>-?\d+\.?\d*) *(?P<unit>.+)?',
-            'curve': r'^curve: ?(?P<curve_id>.+)',
+            'ref_point': r'^(?P<point>(?:x|y)\d)\: *(?P<value>-?\d+\.?\d*) *(?P<unit>.+?)? *(?:(?:@|vs\.?) *(?P<reference>.+))?$',
+            'scale_bar': r'^(?P<axis>x|y)(_scale_bar|sb)\: *(?P<value>-?\d+\.?\d*) *(?P<unit>.+?)?',
```
saraedum, line 868:
```suggestion
            'scale_bar': r'^(?P<axis>x|y)(_scale_bar|sb)\: *(?P<value>-?\d+\.?\d*) *(?P<unit>.+)?',
```
Doesn't adding a `?` here make the unit of scale bar always empty?

svgdigitizer/svgplot.py
```diff
@@ -208,17 +208,14 @@ def units(self):
         ...   </g>
         ... </svg>'''))
         >>> plot = SVGPlot(svg)
-        >>> with TestCase.assertLogs(_) as logs:
-        ...    plot.units
-        ...    print(logs.output)
+        >>> plot.units
         {'x': 'm', 'y': 'A'}
-        ['WARNING:svgplot:Units on y axis do not match. Will ignore unit mA and use A.']
```
saraedum, line 215: I am not sure what happened here. But I think that this was an important test that we do not want to loose.

saraedum (changes requested):
Morally, this adds somewhat CV specific functionality to SVGPlot which supposedly should not know about these things.

I wonder if we should make the `ref_point` pattern a parameter of `SVGPlot.__init__()` and then just expose all the groups that we matched in as a dict coming out of `.marked_points`?

So, for example, when no `ref_point_pattern` is given, we use the old pattern with value & unit. Or even the older pattern without a unit. When we use it for the CV, we initialize `SVGPlot(ref_point_pattern="the pattern in this PR")` and then we magically get `.marked_points[0].unit` and `.marked_points[0].reference`.

I can certainly help quite a bit with the implementation but does that make sense?

Note that this also removes the somewhat odd tuple that `marked_points` returns where you just need to know what the individual entries mean currently.

DunklesArchipel:
In principle we get the x axis unit from `self.svgplot.units['x']`. This can, however, be any kind of string written after a value. Hence, up until this point `V` and `V@RHE` or `V vs. RHE` are considered to be a unit.

Maybe we can use regex on this string in `CV.get_axis_unit(cls, unit):`. Here we would extract the true axis unit as well as the reference.

saraedum:
> In principle we get the x axis unit from `self.svgplot.units['x']`. This can, however, be any kind of string written after a value. Hence, up until this point `V` and `V@RHE` or `V vs. RHE` are considered to be a unit.
> 
> Maybe we can use regex on this string in `CV.get_axis_unit(cls, unit):`. Here we would extract the true axis unit as well as the reference.

Sounds like a great solution to me without complicating things in SVGPlot.

saraedum:
Superseded by #67.

PR #66, opened 2021-09-12
Fix datetime json dump.

### LINKED ISSUE #56: Parse datetime from yaml to dict to json
Apparently when a date is given as `2021-08-15` in the yaml file, it will be shown by python in the dict as  `datetime.date(2021, 8, 15)`. Apparently, `dict` to `json` cannot handle this format.

For now I changed in #51 in the relevant yaml files the date to `2021.08.15`.

DunklesArchipel:
Fixes #56 

Instead of modifying the  `yaml` loader, I changed the way `json` dumps the files. This seemed more intuitive to me:

Yaml NoDateSafeLoader: https://stackoverflow.com/questions/34667108/ignore-dates-and-times-while-parsing-yaml
Json: https://simplernerd.com/python-datetime/


svgdigitizer/__main__.py
```diff
@@ -58,16 +58,22 @@ def cv(svg, sampling_interval, metadata):
...
+    import datetime
+
+    def defaultconverter(o):
+        if isinstance(o, datetime.datetime):
+            return o.__str__()
+
     import json
     with open(Path(svg).with_suffix('.json'), "w") as outfile:
-        json.dump(cv.metadata, outfile)
+        json.dump(cv.metadata, outfile, default=defaultconverter)
```
saraedum, line 76: I don't understand your PR summary. You say that you only change how JSON is written. But actually you do both. So the above, i.e., replace `FullLoader` with `SafeLoader` is not doing enough? Where do these datetimes come from here?

DunklesArchipel: Yes sorry. The `SafeLoader`was my first approach. But that does not solve the problem. I reverted it to FullLoader.

The datetime comes from importing YYYY-MM-DD with yaml (any Loader) with python. In our standard yaml file, we indicate when the file is digitized.

PR #70, opened 2021-09-22
Fixed cv.plot()

DunklesArchipel:
Added a short docu on `def x_label()`

### DIFF
--- a/svgdigitizer/electrochemistry/cv.py
+++ b/svgdigitizer/electrochemistry/cv.py
@@ -75,7 +75,6 @@ class CV():
     def x_label(self):
         r"""
         Return the label on the x-axis of the SVG plot.
-        
         In contrast to the y-label, which only consists of a unit, the x-label consists of a unit and a reference.
         The unit and the reference are united in a single string, which are separated by x_label.
         """
--- a/svgdigitizer/electrochemistry/cv.py
+++ b/svgdigitizer/electrochemistry/cv.py
@@ -74,7 +74,9 @@ class CV():
     @property
     def x_label(self):
         r"""
-        In contrast to the y-label, which only consists of a unit, the x-label consitst of a unit and a reference.
+        Return the label on the x-axis of the SVG plot.
+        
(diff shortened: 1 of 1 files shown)


svgdigitizer/electrochemistry/cv.py
```diff
@@ -183,7 +187,7 @@ def _add_time_axis(self, df):
     def plot(self):
```
saraedum, line 187: Do we want to run plot as a doctest here to see that it works now?

svgdigitizer/electrochemistry/cv.py
```diff
@@ -73,6 +73,10 @@ def get_axis_unit(cls, unit):
 
     @property
     def x_label(self):
+        r"""
+        In contrast to the y-label, which only consists of a unit, the x-label consitst of a unit and a reference.
```
saraedum, line 77:
```suggestion
        Return the label on the x-axis of the SVG plot.
        
        In contrast to the y-label, which only consists of a unit, the x-label consists of a unit and a reference.
```

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: see #71

svgdigitizer/electrochemistry/cv.py
saraedum: (there was a typo.)

saraedum (approved):

svgdigitizer/electrochemistry/cv.py
```diff
@@ -73,6 +73,12 @@ def get_axis_unit(cls, unit):
 
     @property
     def x_label(self):
+        r"""
+        Return the label on the x-axis of the SVG plot.
+
```
DunklesArchipel, line 78:
```suggestion

```

saraedum: (my fault, sorry.)

PR #74, opened 2021-09-23
Generate documentation with Sphinx

saraedum:
With these changes documentation is automatically generated with Sphinx and then uploaded to GitHub Pages.

Currently, most pages contain just a bunch of *todos* and still need to be written:

image

The API documentation is still lacking some introductory text:

image

Such introductory text should be written at the very top of the actual .py file.


saraedum:
Should we merge this as is and then resolve the TODOs in separate Pull Requests? Or should we fix the documentation completely here before merging?

Maybe merging first is better so we can already see the result on GitHub pages.

PR #79, opened 2021-09-30
Added datapackage option in cv (Fix #72)

### LINKED ISSUE #72: Add export functionality to DataPackage?
For compatibility with DataPackage: I think we could/should have a def export_to_datapackage as a functionality of cv. If we just export json identical to yaml we need to create another json for datapackage descriptor.

I have a running version which does according task: export .json, and .csv but fullfilling datapackage criteria and keeping all the yaml information. 

We could also keep datapackage in website and then transform initial .json to datapackage descriptor json. Seems too complicated??!

Shall I pull request?

DunklesArchipel:
Extend cv to create a datapackage consisting of an svg and a json.
`svgdigitizer cv file.svg --metadata file.yaml --package`

Remaining issues:
- [x] the datetime from the original yaml is exported as `null` in the json.
- [x] add test in test.yml


svgdigitizer/__main__.py
```diff
@@ -51,9 +51,11 @@ def digitize(svg, sampling_interval):
 @click.command()
 @click.option('--sampling_interval', type=float, default=None, help=help_sampling)
 @click.option('--metadata', type=click.File("rb"), default=None, help='yaml file with metadata')
+@click.option('--package', is_flag=True, help='create a datapacakge containing a csv and json')
```
saraedum, line 54:
```suggestion
@click.option('--package', is_flag=True, help='create a data packagage containing a csv and json')
```

PR #93, opened 2021-10-19
Fixes #85 and added comment property

### LINKED ISSUE #85: cv.metdata: Metadata is not a property
The following is probably not a property, since it has additonal arguments.
```python
@property
def metadata(self, comment=''):
    ...
```

Once fixed the following can be added to the doctest:

'''python
            A comment describing specifc aspects of the plot can be added, for example to indicate that the original data is rather noisy.::

            >>> cv.metadata(comment='Noisy data')
            {'figure description': {'type': 'digitized', 'measurement type': 'CV', 'scan rate': {'value': 50.0, 'unit': 'V / s'}, 'potential scale': {'unit': 'mV', 'reference': 'RHE'}, 'current': {'unit': 'uA / cm2'}, 'comment': 'Noisy data'}}
'''

DunklesArchipel:
Made metadata a property.

I think it would be more helpful if the comment is included in the SVG file.

TODO:
- [ ] Add test for comment
- [ ] Adapt workflow.md
- [ ] Add comment on echemdb/website on the overview page for a specific system:
      Comment from the curator [Curator Name](Link to ORCID)
      {{ entry.figure_description.comment }}

### DIFF
--- a/svgdigitizer/electrochemistry/cv.py
+++ b/svgdigitizer/electrochemistry/cv.py
@@ -599,8 +599,57 @@ class CV():
         plt.ylabel(self.axis_properties['y']['dimension'] + ' [' + str(self.axis_properties['y']['unit']) + ']')
 
     @property
-    def metadata(self, comment=''):
-        # TODO: Metadata is not a property (see issue #85)
+    @cache
+    def comment(self):
+        r"""
+        Return a comment describing the plot the plot.
+
+        The comment is read from a ``<text>`` in the SVG file such as ``<text>comment: noisy data</text>``.
+
+        EXAMPLES::
+
+            >>> from svgdigitizer.svg import SVG
+            >>> from svgdigitizer.svgplot import SVGPlot
+            >>> from svgdigitizer.electrochemistry.cv import CV
+            >>> from io import StringIO
+            >>> svg = SVG(StringIO(r'''
+            ... <svg>
+            ...   <g>
+            ...     <path d="M 0 200 L 0 100" />
+            ...     <text x="0" y="200">x1: 0 cm</text>
+            ...   </g>
+            ...   <g>
+            ...     <path d="M 100 200 L 100 100" />
+            ...     <text x="100" y="200">x2: 1cm</text>
+            ...   </g>
+            ...   <g>
+            ...     <path d="M -100 100 L 0 100" />
+            ...     <text x="-100" y="100">y1: 0</text>
+            ...   </g>
+            ...   <g>
+            ...     <path d="M -100 0 L 0 0" />
+            ...     <text x="-100" y="0">y2: 1 A</text>
+            ...   </g>
+            ...   <text x="-200" y="330">scan rate: 50 V/s</text>
+            ...   <text x="-400" y="430">comment: noisy data</text>
+            ... </svg>'''))
+            >>> cv = CV(SVGPlot(svg))
+            >>> cv.comment
+            'noisy data'
+
+        """
+        comments = self.svgplot.svg.get_texts('(?:comment): (?P<value>.*)')
+        # TODO: assert that only one label contains the scan rate (see issue #58)
+        # TODO: assert that a rate is available at all (see issue #58)
+
+        # Convert to astropy unit
+        # comments[0].unit = CV.get_axis_unit(rates[0].unit)
+        # rate = float(rates[0].value) * rates[0].unit
+
+        return comments[0].value
+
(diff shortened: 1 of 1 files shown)


svgdigitizer/electrochemistry/cv.py
```diff
@@ -599,8 +602,87 @@ def plot(self):
         plt.ylabel(self.axis_properties['y']['dimension'] + ' [' + str(self.axis_properties['y']['unit']) + ']')
 
     @property
-    def metadata(self, comment=''):
-        # TODO: Metadata is not a property (see issue #85)
+    @cache
+    def comment(self):
+        r"""
+        Return a comment describing the plot the plot.
```
saraedum, line 608:
```suggestion
        Return a comment describing the plot.
```

svgdigitizer/electrochemistry/cv.py
```diff
@@ -599,8 +602,87 @@ def plot(self):
...
+            ...     <text x="-100" y="0">y2: 1 A</text>
+            ...   </g>
+            ...   <text x="-200" y="330">scan rate: 50 V/s</text>
+            ...   <text x="-400" y="430">comment: noisy data</text>
+            ... </svg>'''))
+            >>> cv = CV(SVGPlot(svg))
+            >>> cv.comment
+            'noisy data'
+
+        The seconmd example does not contain a comment::
```
saraedum, line 645:
```suggestion
        This example does not contain a comment::
```

There was a typo here. Also, the "first" above confused me because I thought there was more than one in the first section.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -599,8 +602,87 @@ def plot(self):
...
+
+        The comment is read from a ``<text>`` in the SVG file such as ``<text>comment: noisy data</text>``.
+
+        EXAMPLES:
+
+        The first example contains a comment::
+
+            >>> from svgdigitizer.svg import SVG
+            >>> from svgdigitizer.svgplot import SVGPlot
+            >>> from svgdigitizer.electrochemistry.cv import CV
```
saraedum, line 618:
```suggestion
```

Should work without this line.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -599,8 +602,87 @@ def plot(self):
...
+            ... </svg>'''))
+            >>> cv = CV(SVGPlot(svg))
+            >>> cv.comment
+            'noisy data'
+
+        The seconmd example does not contain a comment::
+
+            >>> from svgdigitizer.svg import SVG
+            >>> from svgdigitizer.svgplot import SVGPlot
+            >>> from svgdigitizer.electrochemistry.cv import CV
```
saraedum, line 649:
```suggestion
```

Should work without this line as well.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -599,8 +602,87 @@ def plot(self):
...
+            ...     <text x="-100" y="0">y2: 1 A</text>
+            ...   </g>
+            ...   <text x="-200" y="330">scan rate: 50 V/s</text>
+            ... </svg>'''))
+            >>> cv = CV(SVGPlot(svg))
+            >>> cv.comment
+            ''
+
+        """
+        if not self.svgplot.svg.get_texts('(?:comment): (?P<value>.*)'):
```
saraedum, line 676:
```suggestion
        if not self.svgplot.svg.get_texts('(?:comment): +(?P<value>.*)'):
```
Should we strip an arbitrary amount of whitespace?

saraedum: I think it's better to pull this repeated code out:

```
comments = self....get_texts()
if not comments:
  return
elif len(comments) > 1:
  warn
normal handling of comments.
```

svgdigitizer/electrochemistry/cv.py
```diff
@@ -599,8 +602,87 @@ def plot(self):
...
+            return ''
+        else:
+            comments = self.svgplot.svg.get_texts('(?:comment): (?P<value>.*)')
+            if len(comments) > 1:
+                logger.warning(f"More than one comment. Irgnoring all comments except for the first: {comments[0]}.")
```
saraedum, line 681:
```suggestion
                logger.warning(f"More than one comment. Ignoring all comments except for the first: {comments[0]}.")
```

saraedum (approved):

PR #97, opened 2021-11-19
Fix svgdigitizer cv when invoked on files in the current directory

saraedum:
without this change, there is an error that the directory `""` cannot be found.


DunklesArchipel (approved):
Great. Thanks for adding the doctest. Now, this `TemporaryData` becomes a bit more clear to me.

PR #98, opened 2021-11-19
Fix cv command when not sampling but unit is not volts

saraedum:
Please merge #97 first before reviewing this PR.


DunklesArchipel (commented):
there might be still an issue with V and mV

PR #94, opened 2021-11-10
add pH estimation

linuxrider:

PR #103, opened 2021-11-22
__main__.cv: Remove all empty keys from the dict (yaml) when loaded.

DunklesArchipel:
[Some Files](https://github.com/echemdb/website/pull/50/files#diff-2e899eb28078fb575ee4846a03283a4911ad43be540e27f5b059c5f055f1ec3f) added to the Website have a large number of keys without value. In the case of the Electrolyte components, this is an issue for #94 . 

Here I simply remove all keys without value from the dictionary.

The question is if we really want to remove all those keys. It might be more simple to add certain values later on if the keys are left as a placeholder.


svgdigitizer/__main__.py
```diff
@@ -174,6 +174,8 @@ def cv(svg, sampling_interval, metadata, package, outdir):
 
     if metadata:
         metadata = yaml.load(metadata, Loader=yaml.SafeLoader)
+        # remove all empty keys
+        metadata = {key: value for key, value in metadata.items() if value}
```
saraedum, line 178: Should we do this recursively? I.e., should
```yaml
a:
  b:
```
also collapse to nothing?

saraedum:
> The question is if we really want to remove all those keys. It might be more simple to add certain values later on if the keys are left as a placeholder.

I think it's easy to fill in defaults later if things are missing. In particular if we have a schema that defines the defaults eventually.

saraedum:
Actually, I wonder whether we should just drop all the empty and null entries already from the yaml completely. If we have a comment in the first line that says, `# See URL for all possible entries and examples` it's probably better than having `unit: # M [one of ... but unfortunately, this list is outdated]` since we won't ever update such comments in legacy data.

DunklesArchipel:
I close this PR, as we agree to remove all empty keys and comments from the YAML prior to submission to echemdb.

PR #114, opened 2022-01-17
Fix lint in svgdigitizer

### LINKED ISSUE #83: Use a proper linter in GitHub Workflow
Currently, we are only performing some PEP8 style checks in the pycodestyle workflow.

However, in the past we had some type issues in untested code paths.

We should use a proper linter to find such errors, e.g., https://github.com/sunnysid3up/python-linter.

saraedum:
Fixes #83.


DunklesArchipel (approved):
looks great. thanks.

Issue #95, opened 2021-11-16
Remove automatic fixing of unit spellings


saraedum:
Eventually we won't be able to fix all unit spellings to make them match the astropy standard.

We should probably just drop this feature and instead produce a good error message when the unit cannot be understood, i.e., explain to the user in two sentences how units have to be spelled and link to a more detailed documentation on astropy.

linuxrider:
#127

linuxrider:
Units can now be written as e.g. "A cm-2", "A cm⁻²", "A / cm2" or "A / cm²" only relying on astropy.

PR #126, opened 2022-01-24
reduce amount unit typos by normalizing superscript chars first

linuxrider:
One of the units in the sato_2006 paper had a missing unit normalization (`10 mV s⁻¹`).
Took this opportunity to reduce the amount of manually specified typos by normalizing superscript first.

### DIFF
--- a/svgdigitizer/electrochemistry/cv.py
+++ b/svgdigitizer/electrochemistry/cv.py
@@ -234,6 +234,30 @@ class CV:
             },
         }
 
+    @staticmethod
+    def to_normal_script(with_superscript):
+        r"""
+        Return the a string with super script characters converted to normal script.
+
+        EXAMPLES::
+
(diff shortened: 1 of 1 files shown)


DunklesArchipel:
In principle, we want to enforce that the units are provided in the style of `astropy` units. Then the unit typoes should be removed completely. (see #95)

In the latest PR on the website, I adapted all units in the SVG files accordingly. Also, the documentation for the digitization of the CVs contains already information on how to indicate the units properly.

Maybe we can use this PR to remove the typos instead. :)

linuxrider:
I tested most of unit strings  found in typos list. With astropy 5 they are already handled correctly.
What seems to be not working is e.g. "milliV" so a mix of single character and word fragment.
"milliampere" works.
I think we could omit this function as well as all the typo stuff.

PR #135, opened 2022-01-27
automatic axis orientation detection

linuxrider:
Checklist


linuxrider:
@saraedum isort is putting an import above the license stuff. This looks strange to me.
Should we do something about it?

svgdigitizer/svgplot.py
```diff
@@ -132,13 +134,26 @@
...
 
 import pandas as pd
 
 logger = logging.getLogger("svgplot")
 
 
+class PageOrientation(Enum):
+    r"""Enum representing the page orientation to be able to to autodetect axis orientation."""
+    PORTRAIT = auto()
+    LANDSCAPE = auto()
```
linuxrider, line 148:
```suggestion
    PORTRAIT = "portrait"
    LANDSCAPE = "landscape"
```

svgdigitizer/svgplot.py
```diff
@@ -132,13 +134,26 @@
...
+
+
+class AxisOrientation(Enum):
+    r"""Enum representing axis orientation."""
+    HORIZONTAL = auto()
+    VERTICAL = auto()
```
linuxrider, line 154:
```suggestion
    HORIZONTAL = "horizontal"
    VERTICAL = "vertical"
```

svgdigitizer/svgplot.py
```diff
@@ -110,6 +110,8 @@
     10  1.0  1.0
 
 """
+
+import itertools
```
linuxrider, line 114:
```suggestion
```

svgdigitizer/svgplot.py
```diff
@@ -336,10 +570,35 @@ def axis_label(axis):
...
+            axis_symbol: axis_label(axis_symbol) for axis_symbol in self.axis_symbols
         }
 
+    @property
+    @cache
+    def _grouped_ref_points(self):
+        r"""
+        Return the reference points grouped by axis symbol.
+        """
+
```
linuxrider, line 582:
```suggestion
         import itertools
```

svgdigitizer/svgplot.py
```diff
@@ -336,10 +569,35 @@ def axis_label(axis):
...
+            axis_symbol: axis_label(axis_symbol) for axis_symbol in self.axis_symbols
         }
 
+    @property
+    @cache
+    def _grouped_ref_points(self):
+        r"""
+        Return the reference points grouped by axis symbol.
+        """
+         import itertools
```
linuxrider, line 581:
```suggestion
        import itertools
```

svgdigitizer/svgplot.py
```diff
@@ -132,13 +133,26 @@
 #  along with svgdigitizer. If not, see <https://www.gnu.org/licenses/>.
 # ********************************************************************
 import logging
+from enum import Enum, auto
```
linuxrider, line 136:
```suggestion
from enum import Enum
```

linuxrider:
So it seems that my locally installed black thinks differently than the one of the github actions.

saraedum:
> So it seems that my locally installed black thinks differently than the one of the github actions.

Are you using conda? If so, you can force the exact same version in your local installation.

svgdigitizer/svgplot.py
```diff
@@ -204,22 +220,182 @@ class SVGPlot:
...
+            >>> plot = SVGPlot(svg)
+            >>> plot.xlabel
+            'E'
+
+            >>> plot = SVGPlot(svg, page_orientation=PageOrientation.LANDSCAPE)
+            >>> plot.xlabel
+            'j'
+
+        """
+
```
saraedum, line 272: Does black really force these empty lines on us? That's weird.

svgdigitizer/svgplot.py
```diff
@@ -131,14 +132,29 @@
...
 from functools import cache
+from itertools import groupby
 
 import pandas as pd
 
 logger = logging.getLogger("svgplot")
 
 
+class PageOrientation(Enum):
+    r"""Enum representing the page orientation to be able to to autodetect axis orientation."""
```
saraedum, line 147:
```suggestion
    r"""
    Page orientation of the source document.
    Used to auto-detect axis orientation.
    """
```

I don't really understand why this is needed. Do we allow inputs to be actually rotated here?

svgdigitizer/svgplot.py
```diff
@@ -131,14 +132,29 @@
...
+    PORTRAIT = "portrait"
+    LANDSCAPE = "landscape"
+
+
+class AxisOrientation(Enum):
+    r"""Enum representing axis orientation."""
```
saraedum, line 153:
```suggestion
    r"""
    The orientation of a plot axis.
    """ 
```

Just trying to rewrite this so it does not exactly replicate what is written in the very line above it.

svgdigitizer/svgplot.py
```diff
@@ -204,22 +220,182 @@ class SVGPlot:
...
-        ylabel=None,
+        page_orientation=PageOrientation.PORTRAIT,
         sampling_interval=None,
         curve=None,
         algorithm="axis-aligned",
     ):
         self.svg = svg
-        self.xlabel = xlabel or "x"
-        self.ylabel = ylabel or "y"
+        self.page_orientation = page_orientation
```
saraedum, line 229: Why is this public? Should this start with a `_`?

svgdigitizer/svgplot.py
```diff
@@ -204,22 +220,182 @@ class SVGPlot:
...
+                    endpoints[0][1] - endpoints[1][1],
+                )
+                delta_x = abs(scalebar[0])
+                delta_y = abs(scalebar[1])
+            # invert on landscape page orientation
+            if (delta_y < delta_x) ^ (
+                self.page_orientation == PageOrientation.LANDSCAPE
+            ):
+                axis_orientations[axis] = AxisOrientation.HORIZONTAL
+            elif (delta_y > delta_x) ^ (
```
saraedum, line 389: What if they are the same `==`?

svgdigitizer/svgplot.py
```diff
@@ -204,22 +220,182 @@ class SVGPlot:
...
+        for axis, orientation in self.axis_orientations.items():
+            if orientation == AxisOrientation.HORIZONTAL:
+                return axis
+        return "x"
+
+    @property
+    @cache
+    def ylabel(self):
+        r"""
+        Return the label of the x axis.
```
saraedum, line 282: The docstring and example always use the `x` axis.

svgdigitizer/svgplot.py
```diff
@@ -204,22 +220,182 @@ class SVGPlot:
...
+
+            >>> plot = SVGPlot(svg, page_orientation=PageOrientation.LANDSCAPE)
+            >>> plot.xlabel
+            'j'
+
+        """
+
+        for axis, orientation in self.axis_orientations.items():
+            if orientation == AxisOrientation.HORIZONTAL:
+                return axis
```
saraedum, line 275:
```suggestion
        [axis] = [axis for axis in self.axis_orientations if self.axis_orientations[axis] == AxisOrientation.HORIZONTAL]
        return axis
```

I think this is shorter and has better error handling (reports an error if there is no such axis or more than one.)

svgdigitizer/svgplot.py
```diff
@@ -336,10 +572,35 @@ def axis_label(axis):
...
+    @property
+    @cache
+    def _grouped_ref_points(self):
+        r"""
+        Return the reference points grouped by axis symbol.
+        """
+
+        # group ref_points by axis
+        keyfunc = lambda x: str(x.label).split(":")[0][
+            :-1
```
saraedum, line 587: Why do we drop the last character?

svgdigitizer/svgplot.py
```diff
@@ -336,10 +572,35 @@ def axis_label(axis):
...
+        ]  # allow multi character axis label
+        ref_points = [points[0] for points in self.labeled_paths["ref_point"]]
+        # without sorting only the last contiguous occurences are returned in the grouping dict
+        ref_points = sorted(ref_points, key=keyfunc)
+        # grouping by first letter of the label text
+        grouped_ref_points = {
+            quantity: list(g) for quantity, g in groupby(ref_points, keyfunc)
```
saraedum, line 594:
```suggestion
            variable: list(g) for variable, g in groupby(ref_points, keyfunc)
```

I think it is more generic to call this the variable of the plot.

svgdigitizer/svgplot.py
```diff
@@ -336,10 +572,35 @@ def axis_label(axis):
...
-            self.xlabel: axis_label(self.xlabel),
-            self.ylabel: axis_label(self.ylabel),
+            axis_symbol: axis_label(axis_symbol) for axis_symbol in self.axis_symbols
         }
 
+    @property
+    @cache
+    def _grouped_ref_points(self):
+        r"""
+        Return the reference points grouped by axis symbol.
```
saraedum, line 582: I had to read the code to understand what this does. Maybe you want to add an example here so it is clear what the returned dict looks like.

svgdigitizer/svgplot.py
```diff
@@ -336,10 +572,35 @@ def axis_label(axis):
...
+        ref_points = sorted(ref_points, key=keyfunc)
+        # grouping by first letter of the label text
+        grouped_ref_points = {
+            quantity: list(g) for quantity, g in groupby(ref_points, keyfunc)
+        }
+
+        if len(grouped_ref_points) > 2:
+            raise ValueError(
+                f"Reference points contain more than two axis names ({list(grouped_ref_points.keys())}). At the moment only 2D plots are supported."
+            )
```
saraedum, line 600:
```suggestion
        if len(grouped_ref_points) != 2:
            raise NotImplementedError(
                f"Currently, there must be exactly two axes since we only support 2D plots. However, we found the variables {list(grouped_ref_points.keys())} on the axes."
            )
```

Should be a NotImplementedError since you say that it's not implemented yet.

svgdigitizer/svgplot.py
```diff
@@ -336,10 +572,35 @@ def axis_label(axis):
...
+        """
+
+        # group ref_points by axis
+        keyfunc = lambda x: str(x.label).split(":")[0][
+            :-1
+        ]  # allow multi character axis label
+        ref_points = [points[0] for points in self.labeled_paths["ref_point"]]
+        # without sorting only the last contiguous occurences are returned in the grouping dict
+        ref_points = sorted(ref_points, key=keyfunc)
+        # grouping by first letter of the label text
```
saraedum, line 592: Why by the first letter of the label?

svgdigitizer/svgplot.py
```diff
@@ -336,10 +572,35 @@ def axis_label(axis):
...
+        r"""
+        Return the reference points grouped by axis symbol.
+        """
+
+        # group ref_points by axis
+        keyfunc = lambda x: str(x.label).split(":")[0][
+            :-1
+        ]  # allow multi character axis label
+        ref_points = [points[0] for points in self.labeled_paths["ref_point"]]
+        # without sorting only the last contiguous occurences are returned in the grouping dict
```
saraedum, line 590: I think that itertools is making things more complicated here. Maybe just:

```
def variable(point): return str(x.label).split(":")[0].strip()
variables = set(variable(point) for point in ref_points)
if len(variables) != 2:
    raise ...
grouped_ref_points = { v: [point for point in ref_points if variable(point) == v] for v in variables }
```

svgdigitizer/svgplot.py
```diff
@@ -204,22 +220,182 @@ class SVGPlot:
...
+            {'E': <AxisOrientation.VERTICAL: 'vertical'>, 'j': <AxisOrientation.HORIZONTAL: 'horizontal'>}
+
+        """
+
+        axis_orientations = {}
+
+        for axis, points in self._grouped_ref_points.items():
+            # distinguish between 2 ref points
+            if len(points) == 2:
+                delta_x = abs(points[0].points[0][0] - points[1].points[0][0])
```
saraedum, line 368: Why `points[0].points[0]`. Shouldn't this be `points[0].far`?

saraedum (changes requested):
It feels me that you are replicating a lot of code from `marked_points` here.

Wouldn't it be easier to just guess `xlabel` and `ylabel` and then look at the `transformation` matrix for `algorithm=mark-aligned` and pick the one that looks "less crazy"?

svgdigitizer/svgplot.py
linuxrider: Some whole page figures in old papers are in landscape orientation.

svgdigitizer/svgplot.py
linuxrider: This is removing the number which is not needed for grouping.

svgdigitizer/svgplot.py
linuxrider: This is not true anymore, it is a residue form the first approach.

svgdigitizer/svgplot.py
linuxrider: True.

svgdigitizer/svgplot.py
linuxrider: No it does not, checked it right now.

> > So it seems that my locally installed black thinks differently than the one of the github actions.
> 
> Are you using conda? If so, you can force the exact same version in your local installation.
> 
> We should probably set a version number in environment.yml to make it clear which version of black we are using.

Now, I have setup a dedicated environment from the provided yaml. It seems the work now.

svgdigitizer/svgplot.py
```diff
@@ -204,22 +220,182 @@ class SVGPlot:
...
+            >>> plot = SVGPlot(svg, page_orientation=PageOrientation.LANDSCAPE)
+            >>> plot.xlabel
+            'j'
+
+        """
+
+        for axis, orientation in self.axis_orientations.items():
+            if orientation == AxisOrientation.HORIZONTAL:
+                return axis
+        return "x"
```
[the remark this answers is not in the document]
linuxrider: Removed.

svgdigitizer/svgplot.py
linuxrider: Very nice.

svgdigitizer/svgplot.py
linuxrider: I agree. itertools has been removed.

svgdigitizer/svgplot.py
linuxrider: True.

linuxrider:
> It feels me that you are replicating a lot of code from `marked_points` here.
> 
> Wouldn't it be easier to just guess `xlabel` and `ylabel` and then look at the `transformation` matrix for `algorithm=mark-aligned` and pick the one that does less rotating?
> 
> It feels to me that such an approach would need very few lines of additional code.

How do I determine the amount of rotation?
The following is returned by `transformation` when (1) the right axis labels are passed and (2) when they are reversed.
```
array([[ 0.01,  0.  ,  0.  ],
           [ 0.  , -0.01,  1.  ],
           [ 0.  ,  0.  ,  1.  ]])

array([[ 0.  , -0.01,  1.  ],
           [ 0.01,  0.  ,  0.  ],
           [ 0.  ,  0.  ,  1.  ]])
```

linuxrider:
> > It feels me that you are replicating a lot of code from `marked_points` here.
> > Wouldn't it be easier to just guess `xlabel` and `ylabel` and then look at the `transformation` matrix for `algorithm=mark-aligned` and pick the one that does less rotating?
> > It feels to me that such an approach would need very few lines of additional code.
> 
> How do I determine the amount of rotation? The following is returned by `transformation` when (1) the right axis labels are passed and (2) when they are reversed.
> 
> ```
> array([[ 0.01,  0.  ,  0.  ],
>        [ 0.  , -0.01,  1.  ],
>        [ 0.  ,  0.  ,  1.  ]])
> 
> array([[ 0.  , -0.01,  1.  ],
>        [ 0.01,  0.  ,  0.  ],
>        [ 0.  ,  0.  ,  1.  ]])
> ```

I have no clue if this the correct method to determine the rotation, but I just compared the sums of the diagonals of the matrices.

linuxrider:
With the new approach, the determination of the x and y axis are postponed until it is really needed.
This is necessary since `transformation` needs `marked_points`.

linuxrider:
Since I changed 
```if (
                axis == self.xlabel
                and scalebar[0] < 0
                or axis == self.ylabel
                and scalebar[1] > 0
            ):
                scalebar = (-scalebar[0], -scalebar[1])
```
to
`scalebar = (abs(scalebar[0]), -abs(scalebar[1]))`
I added a test case for a x scale bar. I took the data from scaling_factor test case.
Point x2 was removed and a x scale bar was added with points on x1 and the formed x2 position.

svgdigitizer/svgplot.py
```diff
@@ -197,29 +208,163 @@ class SVGPlot:
...
+            ...     <path d="M -100 0 L 0 0" />
+            ...     <text x="-100" y="0">j2: 1 A</text>
+            ...   </g>
+            ... </svg>'''))
+            >>> plot = SVGPlot(svg)
+            >>> plot.axis_orientations
+            {'E': <AxisOrientation.HORIZONTAL: 'horizontal'>, 'j': <AxisOrientation.VERTICAL: 'vertical'>}
+        """
+
+        from numpy import trace
```
saraedum, line 344: It's a surprisingly reasonable approach to use the trace here. Since you told me that you are just guessing by looking at a few examples, let me try to explain why this is not a bad idea.

The affine transformation matrix is a 3×3 matrix. Essentially consisting of a 2×2 matrix in the top-left corner that describes a linear transformation and a 2×1 matrix in the top right that describes a translation. The rest of the matrix is most of the time just `0 0 1` in the bottom row.

(Not terribly import here, but your approach obviously fails if the bottom row is different in the two transformation matrices. Probably this does not happen here.)

So the question is: Looking at that 2×2 part, how can we estimate the amount of rotation that is happening?

Let's pretend for a moment that we just have a linear transformation and only look at the 2×2 matrix part and let's pretend that it's just a rotation by the angle θ. In that case, it's actually true that the trace is 1 + 2 cosθ. So, given the trace we can recover the absolute value of the angle. And more importantly, a small trace means rotation by a large angle, a small trace rotation by a large angle, at least when these angles are less that ±90°. (And a similar statement is true for angles close to 180° btw.)

The above explains why your approach works when the transformation is only a rotation.

Now what happens when the axes are differently scaled, i.e., when the transformation is not only a rotation matrix but a rotation matrix times a diagonal matrix. I think the approach still works because the scaling is compatible in the two transformations you compute.

I am not actually sure what happens when there is also shear. It might magically work again because the shear is somehow consistent in the two cases but I am not sure.

I think the safe approach here is to decompose the 2×2 matrix into a (rotation with flip) and a shear matrix. This is called a QR decomposition. Of this QR decomposition we can then compare the rotation angle with your approach.

---

So here's what we should do:

Grab the top left 2×2 matrix that describes the linear transformation of the affine transformation. QR decompose this matrix, this gives us a Q matrix which is orthogonal, i.e., a rotation, a reflection, or a combination of both. If the determinant of Q is 1, it's a rotation and so its trace measures the amount of rotation (the bigger the trace, the smaller the rotation.) If the determinant is -1, it's a reflection or a rotation+reflection. We can ignore the case that it's just a reflection since it's just a reflection and a rotation by 0°. So let's say it's a rotation+reflection. Then we undo the reflection by swapping the rows of the matrix Q and compute its trace and again, a large trace means little rotation, a small trace means big rotation.

For this to work, we do not even need the bottom row of the affine transformations to be the same `0 0 c`. It works even if they differ.

saraedum (changes requested):

linuxrider:
@saraedum I think it works now. Maybe you could another look if I implemented your suggestion correctly.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -603,7 +607,9 @@ def _add_voltage_axis(self, df):
         voltage = 1 * CV.get_axis_unit(self.x_label.unit)
         # Convert the axis unit to SI unit V and use the value
         # to convert the potential values in the df to V
-        df["E"] = df["x"] * voltage.to(u.V).value  # pylint: disable=E1101
+        df["E"] = (
+            df[self.svgplot.xlabel] * voltage.to(u.V).value
+        )  # pylint: disable=E1101
```
saraedum, line 612: For me, pylint does not complain about anything here.

svgdigitizer/svgplot.py
```diff
@@ -795,6 +1012,43 @@ def transformation(self):
...
+            ... </svg>'''))
+            >>> SVGPlot(svg, algorithm='mark-aligned').transformation
+            array([[ 0.01,  0.01, -1.  ],
+                   [ 0.  , -0.01,  1.  ],
+                   [ 0.  ,  0.  ,  1.  ]])
+
+        """
+        return self._transformation(self.xlabel, self.ylabel, self._algorithm)
+
+    def _transformation(self, xlabel, ylabel, algorithm):
```
saraedum, line 1046: `algorithm` is only used in some places. In other places you use `self._algorithm`. I'll change this to be a static class method instead.

svgdigitizer/svgplot.py
```diff
@@ -204,22 +215,172 @@ class SVGPlot:
...
+            >>> plot = SVGPlot(svg)
+            >>> plot.axis_orientations
+            {'E': <AxisOrientation.HORIZONTAL: 'horizontal'>, 'j': <AxisOrientation.VERTICAL: 'vertical'>}
+
+        """
+        from numpy import trace
+        from numpy.linalg import det, qr
+
+        axis_orientations = {}
+        case1 = self._transformation(
```
saraedum, line 348: Since you are doing the same thing twice here, it should be possible to simplify this. Let me try.

svgdigitizer/svgplot.py
```diff
@@ -204,22 +215,172 @@ class SVGPlot:
...
+            self.marked_points[f"{self.axis_variables[1]}1"],
+            self.marked_points[f"{self.axis_variables[1]}2"],
+            self.marked_points[f"{self.axis_variables[0]}1"],
+            self.marked_points[f"{self.axis_variables[0]}2"],
+            "mark-aligned",
+        )
+        case2 = [sublist[:-1] for sublist in case2[:-1]]
+        q_2, _ = qr(case2, mode="complete")
+        if det(q_2) < 0:
+            case2 = case2[::-1]
```
saraedum, line 370: This should be `q_2`.

svgdigitizer/svgplot.py
```diff
@@ -204,22 +215,172 @@ class SVGPlot:
...
+
+        if trace(case1) > trace(case2):
```
saraedum, line 372: These should be the `Q` matrices.

svgdigitizer/svgplot.py
saraedum: > Then we undo the reflection by swapping the rows of the matrix Q and compute its trace and again, a large trace means little rotation, a small trace means big rotation.

Sorry, that's nonsense. Such a swap corresponds to a flip with axis x=y.

saraedum:
@linuxrider, I fixed a bug in the algorithm (sorry, my algorithm had a bug) and simplified a few things (hopefully.)

Please have a look :)

saraedum (approved):

linuxrider:
@saraedum I like your changes.

PR #139, opened 2022-02-04
Fix non-orthogonal scalebars

saraedum:


saraedum:
@linuxrider I forget what was the problem here. Is it possibly that this actually works correctly now? Do you have an example where this does not work?

linuxrider:
@DunklesArchipel  @saraedum 
`--skewed` only works when the marked points sit on the axis (i.e. the axis vectors intersect with the plot origin) with no offset like for stacked plots.
Right now we do not account for a possible offset of the vectors. @saraedum is it possible to take the offset into account?

# Examples
```python
from svgdigitizer.svgplot import SVGPlot
from svgdigitizer.svg import SVG


with open("x_and_y_scale_bar_skewed_org.svg","r") as file:
    svg = file.read()
    SVGPlot(SVG(svg),algorithm="mark-aligned").plot()
    
with open("x_and_y_scale_bar_skewed_points.svg","r") as file:
    svg = file.read()
    SVGPlot(SVG(svg),algorithm="mark-aligned").plot()
    
with open("x_and_y_scale_bar_skewed_points_tilted_y.svg","r") as file:
    svg = file.read()
    SVGPlot(SVG(svg),algorithm="mark-aligned").plot()
    
with open("x_and_y_scale_bar_skewed_points_tilted_y_close_x.svg","r") as file:
    svg = file.read()
    SVGPlot(SVG(svg),algorithm="mark-aligned").plot()
    
with open("x_and_y_scale_bar_skewed_points_tilted_y_top_x.svg","r") as file:
    svg = file.read()
    SVGPlot(SVG(svg),algorithm="mark-aligned").plot()
    
with open("x_and_y_scale_bar_skewed_points_tilted_y_same_distance_x_y.svg","r") as file:
    svg = file.read()
    SVGPlot(SVG(svg),algorithm="mark-aligned").plot()
    
with open("x_and_y_scale_bar_skewed_points_tilted_x_y_same_distance_x_y.svg","r") as file:
    svg = file.read()
    SVGPlot(SVG(svg),algorithm="mark-aligned").plot()
    
with open("x_and_y_scale_bar_skewed_points_tilted_x_y_close_x.svg","r") as file:
    svg = file.read()
    SVGPlot(SVG(svg),algorithm="mark-aligned").plot()
```
y scale bar is nonfunctional due to typo. It can be ignored.
- original (with scale bar)
x_and_y_scale_bar_skewed_org

- original with only marked points
x_and_y_scale_bar_skewed_points

- y axis vector tilted and relatively close to the plot origin, x axis vector far away at the bottom
x_and_y_scale_bar_skewed_points_tilted_y

- tilted y axis vector and x axis vector close to origin of plot
x_and_y_scale_bar_skewed_points_tilted_y_close_x

- y axis vector tilted and relatively close to the plot origin, x axis vector far away at the top
x_and_y_scale_bar_skewed_points_tilted_y_top_x

- y axis vector (outside the canvas to the left) tilted with roughly the same distance to the plot origin as the x axis vector
x_and_y_scale_bar_skewed_points_tilted_y_same_distance_x_y

- y axis vector and x axis vector tilted  with roughly the same distance to the plot origin
x_and_y_scale_bar_skewed_points_tilted_x_y_same_distance_x_y

- y axis vector and x axis vector tilted, x vector close to origin
x_and_y_scale_bar_skewed_points_tilted_x_y_close_x

saraedum:
Sorry, but which of the above is supposedly wrong?

linuxrider:
You are right. I should have explained it in more detail.
I numbered the different cases now. 
- 1 and 2 are correct. 
- But as soon as we tilt the y axis vector in case 3 we see that the extracted values for the x axis are extremely positive. 
- To demonstrate what is causing that I moved the x axis vector in 4 close to the origin of the plot (which is in the flatter part of the cyclic voltammogram as you can see in 1). And the x value again show the expected range. So plot 4 is correct.
- In 5 the x axis vector moved to the top and has roughly the same distance to the plot origin as in 3. This time the x values are shifted negatively.
- In 6 the y axis vector is moved far away from the plot origin to the left (unfortunately outside the canvas). And we see that the y values are still ok and the x values dont change as well.
- When the x axis vector is tilted with roughly the same angle as y in 7, both x and y values are off.
- With the x axis vector again close to plot origin in 8 we again get good x values.
Can this be called a projection problem?
The algorthim works correctly as long as only one axis is tilted and the other nontilted axis intersects with the plot origin.
In principle the axis vectors need to be moved so that they intersect the plot origin to get correct results.
I thought about but could not come up with a clever way to achieve this since both axis vectors can have a offset.

saraedum:
>  The algorthim works correctly as long as only one axis is tilted and the other nontilted axis intersects with the plot origin.

I don't fully understand what you understand as "correctly". It seems to me that we provided bad input so we get the correct (but unexpected) output.

linuxrider (approved):

PR #140, opened 2022-02-19
cv: Extract more metadata from SVG

DunklesArchipel:
Checklist
* [x] Added a test for this change. (Updated tests)

Adds the possibility to extract the following metadata from the SVG:
* Figure label, i.e., 2b
* Tags: OER, HER
* The curve identifier

The new axes metadata scheme has been adapted to the frictionless table schema https://specs.frictionlessdata.io/table-schema/
Further TODOs:
- [x] 'data_description.axes' should be adapted to nest a list of axis properties.


.pylintrc
```diff
@@ -4,4 +4,4 @@ max-module-lines=65536
 good-names=x,y,z,A,t,df,cv
 
 [MESSAGES CONTROL]
-disable=import-outside-toplevel,too-many-arguments,logging-fstring-interpolation,use-maxsplit-arg,fixme
+disable=import-outside-toplevel,too-many-arguments,logging-fstring-interpolation,use-maxsplit-arg,fixme,too-many-locals
```
DunklesArchipel, line 7: There is an error in line 201 in __main__, saying there are too-many-locals.
I don't know why this issue did not appear in the previous PRs.

saraedum: I don't think we should disable this check. I pushed a fix to your branch.

doc/news/cvmetadata2.rst
```diff
@@ -0,0 +1,7 @@
+**Added:**
+
+* Added properties to extract `tags` describing the data, `figure_label` and `curve_label` from the SVG, which are included in the property `metadata`.
+
+**Changed:**
+
+* Changed entry `axes` in `figure description` and `data description` to `field` according to `Frictionless Table Schema <https://specs.frictionlessdata.io/table-schema>`_ and update output json in CLI option `cv` according to this scheme.
```
saraedum, line 7:
```suggestion
* Changed entry `axes` in `figure description` and `data description` to `field` according to `Frictionless Table Schema <https://specs.frictionlessdata.io/table-schema>`_.
```

doc/news/cvmetadata2.rst
```diff
@@ -0,0 +1,7 @@
+**Added:**
+
+* Added properties to extract `tags` describing the data, `figure_label` and `curve_label` from the SVG, which are included in the property `metadata`.
```
saraedum, line 3:
```suggestion
* Added property `CV.tags`, `CV.figure_label`, and `CV.curve_label` which are extracted from `<text>` elements in the SVG. The extracted values are stored in `CV.metadata`.
```

I didn't understand what you did until I read the code. Maybe it's easier to understand now?

svgdigitizer/electrochemistry/cv.py
```diff
@@ -312,6 +317,114 @@ def x_label(self):
...
+    @property
+    @cache
+    def figure_label(self):
+        r"""
+        Return the figure name.
+
+        The figure name is read from a ``<text>`` in the SVG file
+        such as ``<text>figure: 2b</text>``.
+
+        Besides `figure`, also `fig` is acceptable in the text field.
```
saraedum, line 329:
```suggestion
        Besides `figure`, also `fig` is acceptable in the text field.
```

I don't think we need that amount of convenience here. From the [Zen of Python](https://en.wikipedia.org/wiki/Zen_of_Python):

> There should be one– and preferably only one –obvious way to do it.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -320,6 +433,8 @@ def rate(self):
 
         The scan rate is read from a ``<text>`` in the SVG file such as ``<text>scan rate: 50 V/s</text>``.
 
+        Besides `scan rate`, also `rate` is acceptable in the text field.
```
saraedum, line 436: It's good that this is pointed out in the documentation now.

Maybe we should only allow "scan rate" here. I don't see why that would be confusing to anyone.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -312,6 +317,114 @@ def x_label(self):
...
+                f"More than one text field with figure labels. Ignoring all text fields except for the first: {figure_labels[0]}."
+            )
+
+        return figure_labels[0].label
+
+    @property
+    @cache
+    def curve_label(self):
+        r"""
+        Return the curve label.
```
saraedum, line 378:
```suggestion
        A descriptive label for this curve to distinguish it from other curves in the same plot.
```

A property should not say "return" imho. And we should try not just to repeat what the method name already says.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -312,6 +317,114 @@ def x_label(self):
             match[0], match[1], match[2] or "unknown"
         )
 
+    @property
+    @cache
+    def figure_label(self):
+        r"""
+        Return the figure name.
```
saraedum, line 324:
```suggestion
        An identifier of the plot to distinguish it from other figures on the same page.
```

svgdigitizer/electrochemistry/cv.py
```diff
@@ -772,11 +887,63 @@ def simultaneous_measurements(self):
...
+            ...   <text x="-400" y="430">tags: BCV, HER, OER</text>
+            ... </svg>'''))
+            >>> cv = CV(SVGPlot(svg))
+            >>> cv.tags
+            ['BCV', 'HER', 'OER']
+
+        """
+        tags = self.svgplot.svg.get_texts("(?:tags): (?P<value>.*)")
+
+        if not tags:
```
saraedum, line 937: I don't think this is necessary. The return at the bottom does the right thing already.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -814,28 +981,39 @@ def metadata(self):
...
+                                  'measurement type': 'CV',
+                                  'fields': [{'name': 't', 'unit': 's'},
+                                  {'name': 'E', 'unit': 'V', 'reference': 'RHE'},
+                                  {'name': 'j', 'unit': 'A / m2'}]}}
 
         """
         metadata = self._metadata.copy()
+        # Add experimental to metadata
+        metadata.setdefault("experimental", {})
+        metadata["experimental"]["tags"] = self.tags
```
saraedum, line 1012: This will override `tags` from the yaml even if `self.tags==[]`. Is that intentional now?

svgdigitizer/electrochemistry/cv.py
```diff
@@ -814,28 +981,39 @@ def metadata(self):
...
+        # Add source to metadata
+        metadata.setdefault("source", {})
+        metadata["source"]["figure"] = self.figure_label
```
saraedum, line 1015: These will override values from the yaml even if `self.*_label==""`.

saraedum (changes requested):

saraedum (approved):

doc/news/cvmetadata2.rst
DunklesArchipel: Looks good to me.

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: Makes sense. I did not think about that.

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: That was more or less what we decided on in the past for other entries. By overriding the keys in the YAML it is clear that only the keys in the SVG file matter.

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: see above

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: This was mostly out of convenience. I agree that it is probably more useful to only have a single option.

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: see below comment on `scan rate`

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: This will actually raise an error:
`IndexError: list index out of range`

svgdigitizer/__main__.py
```diff
@@ -278,6 +275,31 @@ def digitize_cv(svg, sampling_interval, metadata, package, outdir, skewed):
...
+
+    package = Package(cv.metadata, base_path=outdir or os.path.dirname(csvname))
+    package.infer(os.path.basename(csvname))
+    # Update fields describing the data in the CSV
+    package_fields = package.descriptor["resources"][0]["schema"]["fields"]
+    data_description_fields = package.descriptor["data description"]["fields"]
+
+    new_fields = []
+    for idx, field in enumerate(package_fields):
+        if field["name"] == data_description_fields[idx]["name"]:
```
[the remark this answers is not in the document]
DunklesArchipel: This is correct. As far as I understand your approach (and mine), the field items have to be in the same order. I will address this in a separate PR. It might be convenient to use frictionless and its field class.

.pylintrc
DunklesArchipel: 👍

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: fixed

svgdigitizer/electrochemistry/cv.py
saraedum: Sure. But you're overriding them even if there are none in the SVG.

svgdigitizer/__main__.py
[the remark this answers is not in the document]
saraedum: Ok. So I think what you are doing here is wrong ;)

svgdigitizer/electrochemistry/cv.py
saraedum: Sorry, my bad.

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: ok, so you mean that the key should not exist when it is empty?

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: I added an exception since this is actually mandatory.

DunklesArchipel: This will cause some trouble. Actually, it is only mandatory in echemdb. So we should check it there.

svgdigitizer/__main__.py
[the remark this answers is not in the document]
DunklesArchipel: This will be addressed in #142

svgdigitizer/electrochemistry/cv.py
saraedum: Yes, I think that the tags should not exist when they are missing from both the YAML and the SVG.

But when tags are given in the YAML but *not* in the SVG, then they should certainly not be empty.

doc/news/cvmetadata2.rst
```diff
@@ -0,0 +1,7 @@
+**Added:**
+
+* Added property `CV.tags`, `CV.figure_label`, and `CV.curve_label` which are extracted from `<text>` elements in the SVG. The extracted values are stored in `CV.metadata`.
+
+**Changed:**
+
+* Changed entry `axes` in `figure description` and `data description` to `field` according to `Frictionless Table Schema <https://specs.frictionlessdata.io/table-schema>`_.
```
saraedum, line 7:
```suggestion
* Changed entry `axes` in `figure description` and `data description` to `fields` according to `Frictionless Table Schema <https://specs.frictionlessdata.io/table-schema>`_.
```

svgdigitizer/electrochemistry/cv.py
saraedum: Let me try to propose a better implementation here.

DunklesArchipel: I think we should simply overwrite the tags in the yaml. At the moment tags do not yet exist. I currently add them to the existing SVG files in the website repo.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -814,73 +982,85 @@ def metadata(self):
...
-            "value": float(self.rate.value),
-            "unit": str(self.rate.unit),
-        }
-        metadata["figure description"].setdefault("axes", {})
-        metadata["figure description"]["axes"] = {
-            self.axis_properties["x"]["dimension"]: {
-                "unit": str(CV.get_axis_unit(self.x_label.unit)),
-                "reference": self.x_label.reference,
-                "orientation": "x",
+        metadata = {
```
saraedum, line 1010: I find this much easier to read than the previous explicit code.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -312,13 +316,124 @@ def x_label(self):
...
+
+        """
+        figure_labels = self.svgplot.svg.get_texts("(?:figure): (?P<label>.+)")
+
+        if len(figure_labels) > 1:
+            logger.warning(
+                f"More than one text field with figure labels. Ignoring all text fields except for the first: {figure_labels[0]}."
+            )
+
+        if not figure_labels:
```
saraedum, line 366: I rewrote all the properties so they fall back to the yaml metadata when no value has been specified in the SVG. What do you think?

DunklesArchipel: Sure we can do it like that. For the reviewer of the files, it might be a bit awkward since one has to check both the SVG and the YAML to figure out if something is missing.
We can always revoke this change if we find out later that it is not convenient.

saraedum: Sure. Then again, we should make sure that the checks in the website repository alert us automatically if something is missing.

saraedum:
Zulip does not seem to work. I tried to call you but that didn't work either :shrug:

PR #142, opened 2022-03-14
Implement Frictionless and refactor datapackage creation

### LINKED ISSUE #141: Create datapackage with frictionless and improve merging of fields
Is it possible that these don't match? If they don't match, then you end up with fewer fields describing the resource, i.e., `package.descriptor.resources.0.schema.fields` has fewer entries than there are columns in the CSV.

Did you mean to do something like this?
```python
assert all(p["name"] == d["name"] for (p,d) in zip(package.descriptor["resource"][0]["schema"]["fields"], package.descriptor["data description"]["fields"])), "error message"
package.descriptor["resource"][0]["schema"]["fields"] = package.descriptor["data description"]["fields"] = [p | d for (p,d) in as above]
```

_Originally posted by @saraedum in https://github.com/echemdb/svgdigitizer/pull/140#r825478199_

DunklesArchipel:
Checklist
* [x] Added a test for this change. -- Fixed Test instead.

TODOs
- [x] Remove data description from package descriptor

Fixes #141 

----
**Dependencies**
- [x] #140


saraedum (approved):
Awesome :)

PR #143, opened 2022-03-17
Fix removing data_description.fields during datapackage creation

DunklesArchipel:
Checklist
* [x] Added a test for this change. -> Changed a test accordingly.

In #142 erroneously `data_description` was removed completely during the creation of the data packages. Instead, we only want to remove `data_description.fields`, since these are now included in the field description of the data package resource.


linuxrider (approved):
Looks good to me.

PR #125, opened 2022-01-23
svgdigitizer documentation & implementation of myst_nb

DunklesArchipel:
A rough draft with some example files to demonstrate the capabilities of svgdigitizer.

Checklist
* [x] Added an entry in `doc/news/`. Added maindocu.rst 
* [x] Added a test for this change. No test is required for these changes.
* [x] Change U to E on the x-axis in the `mustermann...pdf` and update all figures in `workflow.md`.

TODO:
- [x] Change all notebooks to `md:myst` (see https://github.com/echemdb/website/pull/132)
- [x] implement changes from #145 
----
**Dependencies**  
- [x] #136 
- [x] #135 Adapt axis labels accordingly

PR #145, opened 2022-03-20
cv: Usage of j, I, U and E as axis labels 

DunklesArchipel:
This fixes an issue, where the cv module was not able to read SVGs where the labels were x and y.

In this new version, we force the user to use explicitly a combination of `E` or `U` and `I` or `j`, to create a current-voltage plot. 
It does not matter if the voltage is on the x-axis.

Furthermore I added a `frictionless.schema` property in `svgplot`, from which it would be easier to create a datapackage (#148)
This might also simplify the code in some places, but I did not look into that in detail. This aspect should probably be addressed in another issue.

Checklist
* [x] Adapted test files.

TODO:

- [x] Fix all `cv` tests (change x1 to E1, etc)
- [x] Add doctests to new properties
- [x] Carefully check the description of all former doctests
- [x] Update docu? (see #125)
- [x] Remove test for briega
- [x] Change `rate` to `scan_rate`


svgdigitizer/svgplot.py
```diff
@@ -1656,6 +1656,69 @@ def _min_real_root(cls, polynomial, tmin, tmax):
 
         return min(eligible_roots)
 
+    @property
+    def schema(self):
+        """A frictionless `Schema` object, including a `Fields` object
+        describing the voltage and current axis of the originlal plot
+        including original units. The reference electrode of the
+        potential/voltage axis is also given (if available).
```
DunklesArchipel, line 1664:
```suggestion
        """A frictionless `Schema` object, including a `Fields` object
        describing the voltage and current axis of the original plot
        including original units.
```

doc/.gitignore
```diff
@@ -1 +1,2 @@
 /generated
+.ipynb_checkpoints
```
saraedum, line 2: This file should end with a newline character.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -176,27 +217,45 @@ def axis_properties(self):
...
-        In this second example a current density 'j' is plotted on the y-axis in `uA / cm2`::
+        """
+        dimensions = list(set(["E", "U"]).intersection(self.svgplot.schema.field_names))
+
+        if len(dimensions) == 1:
+            if self.svgplot.schema.get_field(dimensions[0])["orientation"] == "x":
+                return dimensions[0]
+            raise Exception("The voltage must be on the x-axis.")
+
+        raise Exception("No voltage axis or more than one voltage axis found.")
```
DunklesArchipel, line 250: I changed this part such that only voltages are allowed on the x-axis.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -206,72 +265,102 @@ def axis_properties(self):
...
-    @classmethod
-    def get_axis_unit(cls, unit):
-        r"""
-        Return `unit` as an `astropy <https://docs.astropy.org/en/stable/units/>`_ unit.
+        if len(dimensions) == 1:
+            if self.svgplot.schema.get_field(dimensions[0])["orientation"] == "y":
+                return dimensions[0]
+            raise Exception("The current must be on the x-axis.")
+
+        raise Exception("No current axis or more than one current axis found.")
```
DunklesArchipel, line 296: The same as above for the voltage axis, except that currents are only allowed on the y-axis.

doc/news/cv4.rst
```diff
@@ -0,0 +1,15 @@
...
+
+* Added a property `schema` to `svgplot.SVGPlot`, describing the data extracted from the plot.
+* Added a property `figure_schema`to `electrochemistry.cv.CV`, containing the original figure axis properties.
+* Added a property `data_schema`to `electrochemistry.cv.CV`, containing the axis properties of the property `electrochemistry.cv.CV.df`.
+* Added `voltage_dimension` and `current_dimension` to `electrochemistry.cv.CV`, providing the dimension of the voltage and the current.
+
+**Changed:**
+
+* Changed `electrochemistry.cv.CV` such that it only accepts SVGs where the labels on one axis are either `U` or `E` (V, mV, ...) and another axis are `I` (A, mA, ...) or `j` (A / cm2, uA / m2, ...).
+* Changed `rate`` to `scan_rate`` in `electrochemistry.cv.CV`.
```
saraedum, line 11:
```suggestion
* Changed `rate`` to `scan_rate`` in ``electrochemistry.cv.CV`.
```

svgdigitizer/electrochemistry/cv.py
```diff
@@ -206,72 +265,102 @@ def axis_properties(self):
...
 
-        This method normalizes unit names, e.g., it rewrites 'uA cm-2' to 'uA / cm2' which astropy understands.
+    @property
+    def data_schema(self):
+        """A frictionless `Schema` object, including a field object
+        describing the data generated with :meth:`df`.
+        Compared to :meth:`figure_schema` all fields are given in SI units
```
saraedum, line 302:
```suggestion
        Compared to :meth:`figure_schema` all fields are given in SI units.
```

svgdigitizer/electrochemistry/cv.py
```diff
@@ -206,72 +265,102 @@ def axis_properties(self):
...
         """
-        return u.Unit(unit)
+
+        schema = self.figure_schema
+
+        schema.get_field(self.voltage_dimension)["unit"] = "V"
+        del schema.get_field(self.voltage_dimension)["orientation"]
+        if self.current_dimension == "I":
+            schema.get_field(self.current_dimension)["unit"] = "A"
+        if self.current_dimension == "j":
```
saraedum, line 345:
```suggestion
        elif self.current_dimension == "j":
```

Apparently there is no test case where the current dimension is `I`. Maybe we should add one.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -206,72 +265,102 @@ def axis_properties(self):
...
+            if self.svgplot.schema.get_field(dimensions[0])["orientation"] == "y":
+                return dimensions[0]
+            raise Exception("The current must be on the x-axis.")
+
+        raise Exception("No current axis or more than one current axis found.")
 
-        This method normalizes unit names, e.g., it rewrites 'uA cm-2' to 'uA / cm2' which astropy understands.
+    @property
+    def data_schema(self):
+        """A frictionless `Schema` object, including a field object
```
saraedum, line 300:
```suggestion
        r"""
        A frictionless `Schema` object, including a field object
```

svgdigitizer/electrochemistry/cv.py
```diff
@@ -206,72 +265,102 @@ def axis_properties(self):
...
-        r"""
-        Return the label on the x-axis of the SVG plot.
-        Usually the label on an axis only consits of a unit.
-        In the case of electrochemical data the x-label
-        usually consists of a unit and a reference.
-        The unit and the reference are united in a single string,
-        which are separated by ``x_label`` providing access to
-        the unit and the reference.
+    def figure_schema(self):
+        """A frictionless `Schema` object, including a `Fields` object
```
saraedum, line 360:
```suggestion
        r"""
        A frictionless `Schema` object, including a `Fields` object
```

svgdigitizer/electrochemistry/cv.py
```diff
@@ -206,72 +265,102 @@ def axis_properties(self):
...
+        if len(dimensions) == 1:
+            if self.svgplot.schema.get_field(dimensions[0])["orientation"] == "y":
+                return dimensions[0]
+            raise Exception("The current must be on the x-axis.")
+
+        raise Exception("No current axis or more than one current axis found.")
 
-        This method normalizes unit names, e.g., it rewrites 'uA cm-2' to 'uA / cm2' which astropy understands.
+    @property
+    def data_schema(self):
```
saraedum, line 299: Using intersphinx, both `Schema` and `Field` could link to the frictionless documentation. Not sure if it's worth it.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -283,42 +372,63 @@ def x_label(self):
...
-        )
+        schema.get_field(self.voltage_dimension)["unit"] = match[1]
+        schema.get_field(self.voltage_dimension)["reference"] = match[2] or "unknown"
+
+        return schema
+
+    @classmethod
+    def get_axis_unit(cls, unit):
+        r"""
+        Return `unit` as an `astropy <https://docs.astropy.org/en/stable/units/>`_ unit.
```
saraedum, line 414: If we end up using intersphinx, we could also link to astropy in a stable way here.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -283,42 +372,63 @@ def x_label(self):
...
+
+        This method normalizes unit names, e.g., it rewrites 'uA cm-2' to 'uA / cm2'
+        which astropy understands.
+
```
saraedum, line 418:
```suggestion
```

That's not really true, so maybe better not mention that part and it's probably obvious from the examples.

svgdigitizer/electrochemistry/cv.py
```diff
@@ -494,13 +605,15 @@ def rate(self):
...
         # TODO: Add a more meaningful curve that reflects the shape of a cyclic voltammogram and which is displayed in the documentation (see issue #73).
         r"""
-        Return a dataframe with axis 't', 'E', and 'I' (or 'j).
+        Return a dataframe with axis 't', 'E' (or 'U'), and 'I' (or 'j).
         The dimensions are in SI units 's', 'V' and 'A' (or 'A / m2').
 
-        The dataframe is constructed from the 'x' and 'y' axis of 'svgplot.df',
-        which are usually not in SI units.
+        The dataframe is constructed based on the units and values,
+        determined from ``svplot``. These are usually not in SI units
```
saraedum, line 612:
```suggestion
        determined from ``svgplot``. These are usually not in SI units
```

svgdigitizer/svgplot.py
```diff
@@ -1656,6 +1656,68 @@ def _min_real_root(cls, polynomial, tmin, tmax):
 
         return min(eligible_roots)
 
+    @property
+    def schema(self):
+        """A frictionless `Schema` object, including a `Fields` object
+        describing the voltage and current axis of the original plot
+        including original units.
```
saraedum, line 1663: There's no voltage & current here.

svgdigitizer/svgplot.py
```diff
@@ -1656,6 +1656,68 @@ def _min_real_root(cls, polynomial, tmin, tmax):
...
+            field = {
+                "name": label,
+                "unit": self.axis_labels[label],
+                "orientation": orientation(key.value),
+            }
+            fields.append(field)
+
+        from frictionless import Schema
+
+        return Schema(fields=fields)
```
saraedum, line 1719:
```suggestion
        from frictionless import Schema

        orientations = {
            "vertical": "y",
            "horizontal": "x",
        }

        return Schema(fields={
            "name": label,
            "unit": self.axis_labels[label],
            "orientation": orientations[key.value],
        } for key, label in self.axis_orientations.items())
```

I think you can just do a dict lookup instead of implementing a function. A key error is going to be slightly more cryptic but it cannot happen unless we have a bug in the program anyway.

saraedum (approved):

svgdigitizer/svgplot.py
DunklesArchipel: That's when you copy-paste without paying attention to the actual content... 🙈

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: I assume a doctest is sufficient.

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: I wonder if we need this function anyway. I guess we can simply import astropy where needed. I think this function had more functionality in the past. :)

svgdigitizer/electrochemistry/cv.py
DunklesArchipel: see #151 .
TODOs were added to the code

DunklesArchipel:
I had to fix the version number of click to 8.0.3, since 8.1.0 is responsible for the failed doctests in the previous test runs.

Issue #149, opened 2022-03-29
Bug: Auto-detect axis orientation


DunklesArchipel:
With the new implementations, some of the files on echemdb/website can not be converted anymore, since the axis are not detected properly ([Examples](https://github.com/echemdb/website/tree/main/literature/mello_2018_bromide_18562)).

Here are two files, containing an identical dataset. Some of the axes' labels are placed in different positions.
[bugged.svg](https://user-images.githubusercontent.com/39313137/160552093-957ae119-eb96-4661-a210-84733af5d45f.svg)
[working.svg](https://user-images.githubusercontent.com/39313137/160552096-c4df669e-5065-422f-86ac-3c50c948b0ae.svg)

If you use the current version of svgdigitizer the following different errors appear.
* working: `svgdigitizer cv working.svg`
```python
    current = 1 * CV.get_axis_unit(self.svgplot.axis_labels["y"])
KeyError: 'y'
```
This bug is fixed in #145 

* bugged: `svgdigitizer cv bugged.svg`
```python
 raise UnitConversionError(
astropy.units.core.UnitConversionError: 'uA / cm2' (electrical current density) and 'V' (electrical potential) are not convertible
(echemdb)
```
This is an indication that the x-axis was inferred from j1 and j2 (placed on the y-axis in the SVG)

If you use the new implementation of #145 the conversion works since it does not care whether x is in V or y is in V. Here we observe that the axes were wrongly extracted from the bugged version. In this case, sampling does not work, since digitize_cv tries to sample j.
```json
		"fields": [
			{
				"name": "j",
				"unit": "uA / cm2",
				"orientation": "x"
			},
			{
				"name": "E",
				"unit": "V",
				"orientation": "y",
				"reference": "RHE"
			}
```
in the working version this seems to be fine:
```json
		"fields": [
			{
				"name": "E",
				"unit": "V",
				"orientation": "x",
				"reference": "RHE"
			},
			{
				"name": "j",
				"unit": "uA / cm2",
				"orientation": "y"
			}
```

Comparing the content of the SVG in plain text, I cannot find any differences. Considering that I somehow managed to create a working version, I assume that it has something to to with the positioning of the axis labels or the lines connected to these labels. These are slightly different between both versions (compare position of j labels and the angle of the lines with respect to the plot).

DunklesArchipel:
It can be ruled out that the order of the groups in the file has an influence.

DunklesArchipel:
[...]

PR #153, opened 2022-03-31
Fix bug in axes orientation (see #149)

DunklesArchipel:
Checklist
* [x] No news.
* [x] Add test for `axes_orientation`

This is a fix for a bug  (see #149) in axes orientation which would swap x- and y-axis, if the lines connecting the label with the plot have a transform.

I did not add a piece of news since this bug does not exist in v 0.4.0, hence it is part of the implementation of 0.5.0

@saraedum suggested some code changes in #149, which are unclear to me.


saraedum:
I think we should add a case that was broken before and works now.

saraedum:
Does this actually fix everything in the website repository? @DunklesArchipel

DunklesArchipel:
Added a test for axes orientation. This will fail for svgdigitizer versions <=0.4.0.
In the SVG file some of the lines of the axes labels have a transform, which would flip the x and y axis. (see #149):

saraedum:
> In the SVG file some of the lines of the axes labels have a transform, which would flip the x and y axis. (see https://github.com/echemdb/svgdigitizer/issues/149):

I can't imagine that this makes any difference. If it does, then there is another problem here.

saraedum (changes requested):
Sorry, this change should not work. My original idea that there was something wrong with the scaling of `µA/cm²` to `A/m²` can't really be true.

saraedum (approved):
In the example test case you addded, the values on one axis differ by 50 and on the other axis by .8. So there is roughly a factor 100 in difference here.

svgdigitizer/svgplot.py
```diff
@@ -347,10 +347,13 @@ def score(horizontal, vertical):
             from numpy.linalg import qr
 
             A = self._transformation(
-                self.marked_points[f"{horizontal}1"],
-                self.marked_points[f"{horizontal}2"],
-                self.marked_points[f"{vertical}1"],
-                self.marked_points[f"{vertical}2"],
+                # We do not use the actual label values here but pretend that the marks delimit the unit square [0, 1]×[0, 1].
+                # The actual values might be scaled very differently, e.g., one axis being in μA and one axis in V.
+                # This leads to one axis (and its errors) being rescaled very heavily which can lead to problems in the QR decomposition, see #149.
```
saraedum, line 352:
```suggestion
                # This leads to one axis (and its errors) influencing the transformation matrix too much which can lead to problems in the QR decomposition, see #149.
```

PR #155, opened 2022-05-11
Fixes the python version to 3.9 in setup.py and environment.yaml

DunklesArchipel:
Checklist
* [x] Added a test for this change. -> Not required

Python 3.9 is required to run the module. See also https://github.com/echemdb/echemdb/issues/36.


DunklesArchipel:
In conda forge we apparently took care of this issue https://github.com/conda-forge/svgdigitizer-feedstock/blob/main/recipe/meta.yaml

saraedum (approved):

PR #163, opened 2022-07-26
Add flag `--citation` to include a bibtext string as descriptor in the datapackage

DunklesArchipel:
Checklist

----
The intention of this PR is to include the content of a bibfile as a bibtex string as a descriptor in the datapackage.
While it makes sense that it is stored as a descriptor, it will also simplify the code of echemdb, since the bibliography can now be called as a Descriptor from the datapackage.


svgdigitizer/__main__.py
```diff
@@ -122,6 +131,26 @@ def _create_svgplot(svg, sampling_interval, skewed):
...
+    try:
+        metadata["source"]["bib"]
+    except KeyError as exc:
+        raise KeyError(
+            "The name of the bibfile must be specified in the metadata in `metadata['source']['bib']`."
+        ) from exc
+
+    bibfile = metadata["source"]["bib"]
+    bibliography = parse_file(f"{bibfile}.bib", bib_format="bibtex")
+    return bibliography.entries[bibfile].to_string("bibtex")
```
DunklesArchipel, line 151: This is intended for use with echemdb metadata. Of course, one could also think about providing a filename of the bib file as an argument to the flag. However, I guess automatisation will become complicated. It is possible that multiple files with different names, share a single BIB file.

saraedum (approved):

PR #170, opened 2023-01-30
Move CLI from `__main__.py` to `entrypoint.py`

### LINKED ISSUE #161: Move __main__.py to entrypoint.py
pytest (since a few versions) ignores doctests in `__main__.py` so we should move the contents of this file elsewhere and just call it from the actual `__main__.py`.

DunklesArchipel:
Checklist

Changes the entry point such that the doctests of the CLI are no longer ignored during testing.

Fixes #161 

Dependencies:
- [x] #169 
- [x] #175


DunklesArchipel:
not sure is `__main__.py`  and `entrypoint.py` should share the same doctest at the beginning of the document.

.github/workflows/doc.yml
```diff
@@ -43,7 +47,7 @@ jobs:
         run: |
           python -m http.server 8880 --directory doc/generated/html &
           sleep 1
-          linkchecker --check-extern http://localhost:8880/svgdigitizer/
+          linkchecker --check-extern http://localhost:8880/svgdigitizer/ --ignore-url="https://repology.org/*"
```
saraedum, line 50: Could you add a comment in the workflow why you want to exclude repology?

DunklesArchipel: This change was added in #169 since the check passes locally but not in the GitHub workflow.

DunklesArchipel: A comment was added.

DunklesArchipel: is this the only change you wanted and would it be ready for merging??

saraedum (approved):

saraedum:
@DunklesArchipel please merge if you agree with the changes here.

PR #171, opened 2023-02-04
Fix dependencies

DunklesArchipel:
Checklist

~~Fix module version differences between setup.py and environment.yaml~~

~~Also include poppler in the setup see https://github.com/conda-forge/svgdigitizer-feedstock/issues/12~~

Fix the alphabetical order of the dependencies. 

From local tests, I did not face any issues using `svgdigitizer paginate`, which were supposed to be related to the missing module poppler.


DunklesArchipel:
@linuxrider  do you know what is going on with this poppler package? (https://github.com/conda-forge/svgdigitizer-feedstock/issues/12)

I tested svgdigitizer in a clean environment via pip, mamba, and `pip install -e .`. In all cases, it was possible to paginate a PDF.

Has anyone else experienced issues with this package?

DunklesArchipel:
It appears that `poppler` is not required as a dependency in setup py. At least in the current version, `svgdigitizer paginate` works when the module is installed via
* `pip install -e .` from the cloned repo
* `pip install svgdigitizer` in a clean environment
* `mamba install svgdigitizer` in a clean environment

saraedum (approved):

PR #175, opened 2023-02-13
Fix exception handling

DunklesArchipel:
Checklist

The new version of `pylint` does not allow raising general exceptions.


saraedum:
I believe that technically none of these are `ValueError`:

> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

The error here is never about the argument. I guess virtually all of these are saying that the data parsed does not contain this information or that the data is inconsistent (duplicate value errors.)

Maybe it's worth our while to create exception classes for this? Something like

```
class DataSufficiencyError(RuntimeError): pass
class DataConsistencyError(RuntimeError): pass
```
(I am not sold on that "sufficiency" term though…)

DunklesArchipel:
How about
```python
class SVGAnnotationError: pass
```
Example:

The following exception:
`raise Exception("No current axis or more than one current axis found.")`

would read:
`raise SVGAnnotation error("No current axis or more than one current axis found in the SVG.")`

svgdigitizer/electrochemistry/cv.py
```diff
@@ -149,7 +149,7 @@ class CV:
 
         >>> cv.metadata  # doctest: +NORMALIZE_WHITESPACE
         {'experimental': {'tags': ['BCV', 'HER', 'OER']},
-         'source': {'figure': '2b', 'curve': '0'},
+         'source': {'figure': '2b', 'curve': 'solid'},
```
DunklesArchipel, line 152: see #176

DunklesArchipel:
@saraedum since I never handled exceptions that accurately it would be great to hear your thoughts about the approach.

I created a new module for the errors, which I would import into the svgdigitizer modules.
Currently, this is only implemented in `svgplot.curve`. (This seems to be a tricky one anyway).
The errors will then contain the doctests for the different Errors and could be removed from the main modules.

I hope this is how it should be done... :)

svgdigitizer/models/exceptions.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class SVGContentError(RuntimeError):
+    """Raised when facing issues with the structure of the SVG input."""
+
+
+class AnnotationError(SVGContentError):
+    """The annotations in the SVG are is missing or incorrect."""
+
+class CurveError(SVGContentError):
```
saraedum, line 31: I would not create these super-specialized exceptions. It's just too much work to maintain this code I fear.

svgdigitizer/models/exceptions.py
```diff
@@ -0,0 +1,101 @@
+r"""
```
saraedum, line 1: I had a file like this in mind, yes. What do you mean when you say `models`? Maybe just make it `svgdigitizer/exceptions.py`?

DunklesArchipel: TBH I stored the file in a folder according to some suggested structures I found online. Intuitively, I would have also put it into the parent directory, while this is probably more reasonable for larger projects.

svgdigitizer/models/exceptions.py
DunklesArchipel: So you mean simply 

```python
class SomeError(RuntimeError):
    """Some description"""
```

And then call the Error via
```python
raise SomeError("Error Message")
```

?

saraedum: Yes.

saraedum:
@DunklesArchipel if the tests pass, then this is ready I think.

saraedum (approved):

PR #185, opened 2023-04-03
Bibliography flag supports conversion without BIB file

DunklesArchipel:
Checklist


doc/news/nobib.rst
```diff
@@ -0,0 +1,3 @@
+**Added:**
+
+* Digitizing with flag `--bibliography` creates a bibliography key, which is empty when no BIB files or no citation key is given in `metadata["source"]["citation key"]`.
```
DunklesArchipel, line 3:
```suggestion
* Digitizing with flag `--bibliography` creates a "bibliography key, which is an empty string when no citation key is given in `metadata["source"]["citation key"]` or no BIB file with the given citation key is available.
```

Issue #187, opened 2023-04-04
Fix output of `byexample` 


DunklesArchipel:
`byexample` creates a cut-off at 80 characters and thus the output of the CLI is not complete in the readme.

_Originally posted by @DunklesArchipel in https://github.com/echemdb/svgdigitizer/pull/182#discussion_r1154789704_

DunklesArchipel:
It also seems that we can not suppress output warnings, which makes implementation of `byexample` even harder in #189

PR #189, opened 2023-04-19
Adapt documentation to version 0.10.0

DunklesArchipel:
Checklist


DunklesArchipel:
Disable `byexample' in the tests (see #187)

doc/cli.md
```diff
@@ -1,6 +1,288 @@
...
+df.plot(x='d', y='height', xlabel='distance [m]', ylabel='height [m]', legend=False)
+```
+
+## `cv`
+
+The `cv` option is designed specifically to digitze [cyclic voltammograms (CVs)](https://en.wikipedia.org/wiki/Cyclic_voltammetry). Overall the command `cv` has the same functionality as the [`figure`](figure) command. The differences are as follows.
+
+* Certain keys in the output metadata are directly related to cyclic voltammetry measurements.
+* The units on the x-axis must be equivalent to volt `U` given in units of `V` and those on the y-axis equivalent to current `I` in units of `A` or current density `j` in units of `A / m2`.
+* The voltage unit can given vs. a reference, such as `V vs. RHE`. In that case the dimension should be `E` instead of `U`.
```
TheOpenBIts, line 178:
```suggestion
* The voltage unit can be given vs. a reference, such as `V vs. RHE`. In that case, the dimension should be `E` instead of `U`.
```

doc/cli.md
```diff
@@ -1,6 +1,288 @@
...
+
+!svgdigitizer cv ./files/mustermann_2021_svgdigitizer_1/mustermann_2021_svgdigitizer_1_f2a_blue.svg --sampling-interval 0.01
+```
+
+(advanced-flags)=
+## Advanced flags
+
+### `--si-units`
+
+The flag `--si-unit` is used by the [`figure`](figure) command and commands that inherit from `figure` such as the `cv` command. The units are converted to SI units, if they are compatible with the [astropy](https://www.astropy.org/) unit package. The values in the CSV are scaled respectively and the new units are provided in the output JSON files.
```
TheOpenBIts, line 212:
```suggestion
The flag `--si-unit` is used by the [`figure`](figure) command and commands that inherit from `figure`, such as the `cv` command. The units are converted to SI units, if they are compatible with the [astropy](https://www.astropy.org/) unit package. The values in the CSV are scaled respectively and the new units are provided in the output JSON files.
```

doc/cli.md
```diff
@@ -1,6 +1,288 @@
...
+
+```{warning}
+In some cases conversion to SI units might not result in the desired output. For example, even thought `V` is considered as an SI unit, astropy might convert the unit to `W / A` or `A Ohm`.
```
TheOpenBIts, line 215:
```suggestion
In some cases conversion to SI units might not result in the desired output. For example, even though `V` is considered as an SI unit, astropy might convert the unit to `W / A` or `A Ohm`.
```

doc/cli.md
```diff
@@ -1,6 +1,288 @@
...
+```
+
+### `--metadata`
+
+The flag `--metadata` allows adding metadata to the resource of the datapackage from a yaml file. It is used by the [`figure`](figure) command and commands that inherit from `figure` such as the `cv` command.
```
TheOpenBIts, line 220:
```suggestion
The flag `--metadata` allows adding metadata to the resource of the datapackage from a yaml file. It is used by the [`figure`](figure) command and commands that inherit from `figure`, such as the `cv` command.
```

TheOpenBIts (commented):
Some spelling corrections in the documentation.

linuxrider (approved):

DunklesArchipel:
We merge this without further review. Feel free to add suggestions as PR or issues. :)

PR #199, opened 2023-10-15
Remove unnecessary dependencies

DunklesArchipel:
Checklist


saraedum:
(It's the same problem in all projects that I am involved with it seems ;) )

DunklesArchipel:
some merges went wrong and the git history seems broken.
created new clean #206

PR #201, opened 2024-01-06
Migrate to pyproject.toml

linuxrider:
Checklist


DunklesArchipel:
@saraedum is this change compatible with rever?

linuxrider:
Is there a difference between `version` and `release`?
```python
$VERSION_BUMP_PATTERNS = [
    ('setup.py', r"    version=", r'    version="$VERSION",'),
    ('doc/conf.py', r"release = ", r"release = '$VERSION'"),
]
```

linuxrider:
@saraedum I tried to make the changes so that it works with rever. I'm not sure if I understood how it's supposed to work, however.

svgdigitizer/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.11.0"
```
saraedum, line 1: I don't think this is the recommended way to this anymore since https://peps.python.org/pep-0396/ has been rejected.

You can do something like the following to determine the version number:
```
>>> from importlib.metadata import version
>>> version('package name')
```

pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+    "pyyaml>=6,<7",
+    "scipy>=1.7,<2",
+    "svg.path>=4.1,<5",
+    "svgpathtools>=1.4,<2",
+    "svgwrite>=1.4,<2",
+]
+dynamic = ["version"]
+
+[tool.setuptools.dynamic]
+version = {attr = "svgdigitizer.__init__.__version__"}
```
saraedum, line 29: I'd just hardcode the version number here and update it with the `$VERSION_BUMP_PATTERNS` in rever.

doc/conf.py
```diff
@@ -1,9 +1,11 @@
...
+
 project = 'svgdigitizer'
-copyright = '2021-2023, the svgdigitizer authors'
+copyright = '2021-2024, the svgdigitizer authors'
 author = 'the svgdigitizer authors'
 
-release = '0.11.0'
-
+version = __version__.rsplit('.',1)[0]
+release = __version__
```
saraedum, line 8: This change is not needed, `$VERSION_BUMP_PATTERNS` takes care of updating this.

pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+[project.scripts]
+svgdigitizer = "svgdigitizer.entrypoint:cli"
+
+[tool.setuptools]
+packages = [
+    "svgdigitizer",
+    "svgdigitizer.electrochemistry",
+    "svgdigitizer.test",
+]
+include-package-data = true
```
saraedum, line 44: Your editor does not seem to be configured to end files with a EOL.

pyproject.toml
```diff
@@ -0,0 +1,44 @@
+[build-system]
+requires = ["setuptools>=61.2"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "svgdigitizer"
+license = {text = "GPL 3.0+"}
```
saraedum, line 7: I believe that this should just be the SPDX identifier:
```suggestion
license = "GPL-3.0-or-later"
```

pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+    "pdf2image>=1.16,<2",
+    "pillow>=10.0.1,<11",
+    "pybtex>=0.24,<0.25",
+    "pyyaml>=6,<7",
+    "scipy>=1.7,<2",
+    "svg.path>=4.1,<5",
+    "svgpathtools>=1.4,<2",
+    "svgwrite>=1.4,<2",
+]
+dynamic = ["version"]
```
saraedum, line 26:
```suggestion
```

saraedum:
> Is there a difference between `version` and `release`?
> 
> ```python
> $VERSION_BUMP_PATTERNS = [
>     ('setup.py', r"    version=", r'    version="$VERSION",'),
>     ('doc/conf.py', r"release = ", r"release = '$VERSION'"),
> ]
> ```

No, sphinx calls it release, and setup.py calls it version but it's the same thing.

pyproject.toml
```diff
@@ -0,0 +1,41 @@
+[build-system]
+requires = ["setuptools>=61.2"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "svgdigitizer"
+version = "0.11.0"
+license = {text = "GPL-3.0-or-later"}
```
linuxrider, line 8: @saraedum We can have either the `text`property here or the `file` property. What do you think?

saraedum: Text is better then a file. In particular, `or later` is not part of the license file (it's just the GPL 3 license.)

I am confused that this typing is needed, since here it seems that just spelling out the SPDX identifier should be enough: https://python-poetry.org/docs/pyproject/#license

linuxrider: This is what I got without the property:
```
  × Getting requirements to build editable did not run successfully.
  │ exit code: 1
  ╰─> [95 lines of output]
      configuration error: `project.license` must be valid exactly by one definition (2 matches found):
      
          - keys:
              'file': {type: string}
            required: ['file']
          - keys:
              'text': {type: string}
            required: ['text']
      
      DESCRIPTION:
          `Project license <https://peps.python.org/pep-0621/#license>`_.
      
      GIVEN VALUE:
          "GPL-3.0-or-later"
      
      OFFENDING RULE: 'oneOf'
      
      DEFINITION:
          {
              "oneOf": [
                  {
                      "properties": {
                          "file": {
                              "type": "string",
                              "$$description": [
                                  "Relative path to the file (UTF-8) which contains the license for the",
                                  "project."
                              ]
                          }
                      },
                      "required": [
                          "file"
                      ]
                  },
                  {
                      "properties": {
                          "text": {
                              "type": "string",
                              "$$description": [
                                  "The license of the project whose meaning is that of the",
                                  "`License field from the core metadata",
                                  "<https://packaging.python.org/specifications/core-metadata/#license>`_."
                              ]
                          }
                      },
                      "required": [
                          "text"
                      ]
                  }
              ]
          }
```

linuxrider: We are building with setuptools. Should we switch to poetry?

saraedum: No, I just assumed that they had the same syntax.
Sorry for the noise. I guess `text` is fine after all.

rever.xsh
```diff
@@ -47,7 +47,7 @@ $ACTIVITIES = [
 ]
 
 $VERSION_BUMP_PATTERNS = [
-    ('setup.py', r"    version=", r'    version="$VERSION",'),
+    ('pyproject.toml', r'version\s*=.*', 'version = "$VERSION"'),
```
saraedum, line 50:
```suggestion
    ('pyproject.toml', r'version =', 'version = "$VERSION"'),
```
Not important but there is no need to match the entire line.

saraedum (approved):

pyproject.toml
saraedum, line 8: > If you are using a standard, well-known license, it is not necessary to use this field. Instead, you should one of the [classifiers](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers) starting with License ::. (As a general rule, it is a good idea to use a standard, well-known license, both to avoid confusion and because some organizations avoid software whose license is unapproved.)

from https://packaging.python.org/en/latest/guides/writing-pyproject-toml/

saraedum (changes requested):

pyproject.toml
linuxrider: True. Thank you.

pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+    "pyyaml>=6,<7",
+    "scipy>=1.7,<2",
+    "svg.path>=4.1,<5",
+    "svgpathtools>=1.4,<2",
+    "svgwrite>=1.4,<2",
+]
+
+[project.readme]
+file = "README.md"
+content-type = "text/markdown"
```
saraedum, line 30:
```suggestion
```

> The README’s format is auto-detected from the extension:  README.md → [GitHub-flavored Markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax),

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/

pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+[project.scripts]
+svgdigitizer = "svgdigitizer.entrypoint:cli"
+
+[tool.setuptools]
+packages = [
+    "svgdigitizer",
+    "svgdigitizer.electrochemistry",
+    "svgdigitizer.test",
+]
+include-package-data = true
```
saraedum, line 41:
```suggestion
```

> # By default, include-package-data is true in pyproject.toml, so you do
> # NOT have to specify this line.
> include-package-data = true

saraedum (approved):

pyproject.toml
```diff
@@ -0,0 +1,39 @@
+[build-system]
+requires = ["setuptools>=61.2"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "svgdigitizer"
+version = "0.11.0"
```
linuxrider, line 7:
```suggestion
version = "0.11.0"
readme = "README.md"
```

pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+    "pybtex>=0.24,<0.25",
+    "pyyaml>=6,<7",
+    "scipy>=1.7,<2",
+    "svg.path>=4.1,<5",
+    "svgpathtools>=1.4,<2",
+    "svgwrite>=1.4,<2",
+]
+
+[project.readme]
+file = "README.md"
```
linuxrider, line 29:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,38 @@
+[build-system]
+requires = ["setuptools>=61.2"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "svgdigitizer"
+version = "0.11.0"
+readme = "README.md"
+classifiers = ["License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)"]
+description = "svgdigitizer is a Python library and command line tool to recover the measured data underlying plots in scientific publications."
```
linuxrider, line 10:
```suggestion
classifiers = ["License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)"]
description = "svgdigitizer is a Python library and command line tool to recover the measured data underlying plots in scientific publications."
readme = "README.md"
```

PR #202, opened 2024-01-06
Allow more recent dependencies and test Python 3.12

linuxrider:
Checklist

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,34 +4,33 @@
 name: svgdigitizer-build
 channels:
   - conda-forge
-  - defaults
 dependencies:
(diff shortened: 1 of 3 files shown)


environment.yml
```diff
@@ -6,7 +6,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - astropy>=5,<6
```
saraedum, line 9: This file contains the dependencies for development. There's usually no benefit to allow old versions here. They should be allowed in the `pyproject.toml` though. (However, we are currently not testing for such old versions, are we?)

PR #204, opened 2024-01-09
Enable testing on macOS again

saraedum:
Checklist
* [x] Updated the documentation. Nothing to say there.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -7,7 +7,7 @@ channels:
 dependencies:
   - astropy>=5,<6
   - black
-  - click=8.0.3
+  - click>=8.1.3,<9
   - frictionless>=5.8.3,<6
   - isort
   - jupytext
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ channels:
   - conda-forge
 dependencies:
   - astropy>=5,<6
-  - black
+  - black>=23,<24
   - click>=8,<9
   - frictionless>=5.8.3,<6
   - isort
@@ -21,7 +21,7 @@ dependencies:
   - pip
   - poppler
   - pybtex>=0.24,<0.25
-  - pylint
+  - pylint>=3,<3.1
   - pytest
   - python
   - pyyaml>=6,<7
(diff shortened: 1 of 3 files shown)


saraedum:
Trying to understand what's the problem on macOS.

environment.yml
```diff
@@ -4,34 +4,33 @@
 name: svgdigitizer-build
 channels:
   - conda-forge
-  - defaults
 dependencies:
   - astropy>=5,<6
-  - black
-  - click=8.0.3
+  - black>=23,<24
+  - click>=8,<9
```
DunklesArchipel, line 10: why pin these versions? I thought we always wanted the most recent version to validate our code. The same applies to `isort` which is not pinned.

environment.yml
```diff
@@ -4,34 +4,33 @@
...
   - pillow>=10.0.1,<11
-  - pip>=21,<22
-  - poppler>=22,<23
+  - pip
+  - poppler
   - pybtex>=0.24,<0.25
-  - pylint>=2.12,<3
-  - pytest>=6.2,<7
-  - python>=3.9,<4
+  - pylint>=3,<3.1
```
DunklesArchipel, line 24: This should probably also not have a version number?

environment.yml
saraedum: black changes how they are formatting things with version upgrades. If we don't pin, the version jumps randomly in some PR and that PR is going to fail for no good reason (just because formatting changed.) That's annoying so it's maybe better to make this version bump explicitly in a separate PR when there is a new black.

Are you also asking about click? I don't know. It's the same pin we have for the project. Probably because we are not compatible with click 9?

environment.yml
saraedum: Same as above. Pylint changes its rules when the version jumps. That can be annoying because when a new pylint gets released, PRs are going to fail because of that version change because the rules changed in pylint.

environment.yml
DunklesArchipel: yes sorry for the click. I selected the wrong lines. :)
thanks for the info on black and pylint.

DunklesArchipel (approved):

PR #210, opened 2024-04-26
Fix `scipy` and `svg.path` dependency issue

DunklesArchipel:
Checklist
* [o] Added a test for this change.
* [o] Updated the documentation.


DunklesArchipel:
@saraedum the conversation in https://github.com/conda-forge/svgdigitizer-feedstock/pull/21 it was not conclusive, but I guess we should include these two modules again.

saraedum (approved):

PR #217, opened 2024-10-23
Fix file links in the usage example documentation.

DunklesArchipel:
see #216

PR #220, opened 2024-11-09
Add pixi setup to replace environment.yml

saraedum:
pixi has a lot of advantages over an environment.yml setup for development (while under the hood just wrapping conda in a more convenient way.) The most important ones for us are probably:

* people do not need to understand conda environments and how and when to update them anymore, things are just installed in the correct versions.
* the package versions are locked and known to work so workflows should not fail randomly anymore

Checklist


DunklesArchipel:
Will this eventually mitigate the workflow issues, such as in #219 or #217 ?

saraedum:
> Will this eventually mitigate the workflow issues, such as in #219 or #217 ?

Yes, workflows shouldn't fail anymore when nothing has changed.

saraedum:
@DunklesArchipel can you try whether this actually works for you? I don't think there was anything else I wanted to do here.
The required checks for the protected branch have to be changed before this gets merged (the names of the tests changed.)

DunklesArchipel:
To get this up and running in gitbash for windows, you have to install pixi (for example using the powershell or the installer) and then add the following lines to your `.bashrc`
```sh
export PATH=$PATH:~/.pixi/bin
eval "$(pixi completion --shell bash)"
```

.github/workflows/test.yml
```diff
@@ -12,147 +12,110 @@ jobs:
...
-        shell: bash -l {0}
+        shell: pixi run bash -e {0}
         run: |
-          svgdigitizer cv test/data/xy_rate_without_metadata_skewed.svg --skewed --si-units
-          git diff --no-index --word-diff -- test/data/xy_rate_without_metadata_skewed.json test/data/xy_rate_without_metadata_skewed.json.expected
-          python -c 'import pandas; pandas.testing.assert_frame_equal(pandas.read_csv("test/data/xy_rate_without_metadata_skewed.csv"), pandas.read_csv("test/data/xy_rate_without_metadata_skewed.csv.expected"))'
-      - name: cli-cv-axes-orientation
-        shell: bash -l {0}
-        run: |
-          svgdigitizer cv test/data/axes_orientation.svg --si-units
```
DunklesArchipel, line 120: Did you ommit this on purpose?

.github/workflows/lint.yml
```diff
@@ -9,30 +9,15 @@ concurrency:
...
       - name: black
-        shell: bash -l {0}
-        run: |
-          black --check --diff svgdigitizer
+        run: pixi run black --diff --check
       - name: isort
-        shell: bash -l {0}
-        run: |
-          isort --check --diff svgdigitizer --line-length 256
+        run: pixi run isort --check --diff
```
DunklesArchipel, line 23: why don't we simply run `pixi run lint`?

DunklesArchipel: and I also don't understand why below in the pixi settings we simply run black in the lint tasks, while here we add additional flags `--diff --check`

DunklesArchipel:
I am not familiar with the pixi files. Apparently, the `pixi.loc` file is essential and will probably be created somehow by pixi. But how and when is it updated?
I found this workflow https://pixi.sh/dev/advanced/updates_github_actions/#using-reusable-workflows
Could that be useful?

DunklesArchipel (approved):

.github/workflows/lint.yml
saraedum: Running these without the flags would just reformat the code. But we want to know whether there is a reformatting missing. So with this, the command returns an error code, when it would have to do something.

saraedum:
> I am not familiar with the pixi files. Apparently, the `pixi.loc` file is essential and will probably be created somehow by pixi. But how and when is it updated? I found this workflow https://pixi.sh/dev/advanced/updates_github_actions/#using-reusable-workflows Could that be useful?

I think it's better for this kind of project to just update the file manually. To recreate the file, you just delete it and run pixi whatever.) I am not opposed to recreating the file with a workflow but it's maybe a bit noisy for such a project that is not updated that frequently.

.github/workflows/test.yml
saraedum: No.

saraedum: Probably a merge that I screwed up. Thanks for checking.

DunklesArchipel (approved):

.github/workflows/test.yml
DunklesArchipel: I added it again

DunklesArchipel:
I guess this can be merged.

PR #231, opened 2025-02-03
preparation for pyodide

linuxrider:
PyMuPDF is not dependent on external tools and can be used in pyodide.
Checklist
* [x] Added a test for this change. Tests already in place.
* [x] Updated the documentation. Nothing to update.


linuxrider:
pymupdf is only available as conda package for linux only.
With changing the pdf handling the speed of paginate is increased significantly. For one tested pdf file the speedup was rough 4.

doc/news/pyodide.rst
```diff
@@ -0,0 +1,3 @@
+**Changed:**
+
+* Changed pdf handling from pdf2image to PyMuPDF which allows to run in pyodide.
```
DunklesArchipel, line 3:
```suggestion
* Changed PDF handling from pdf2image to PyMuPDF, allowing to run in pyodide.
```

svgdigitizer/entrypoint.py
```diff
@@ -600,17 +623,13 @@ def paginate(onlypng, template, pdf, outdir):
...
-        for page in range(len(pages))
-    ]
-
-    for page, png in zip(pages, pngs):
-        page.save(png, "PNG")
+    import pymupdf
 
+    doc = pymupdf.open(pdf)
+    for page_idx, page in enumerate(doc):
+        pix = page.get_pixmap(dpi=600)
```
DunklesArchipel, line 630: what does `pix` refer to? Should we use a less ambiguous value name?

DunklesArchipel (approved):

PR #234, opened 2025-02-21
add template elements to generated svgs

linuxrider:
Checklist


linuxrider:
I added the option `--template` for pagination. It takes either an identifier, e.g. `basic`, or a file path `file:<file path>`. the basic template file should be distributed together with svgdigitizer package. I put it in subfolder assets.
@saraedum what do you think?

DunklesArchipel:
We should also add at some point some documentation in the CLI part

https://echemdb.github.io/svgdigitizer/cli.html

Feel free to create an issue to do that later since this will be a bit experimental at the moment.

DunklesArchipel (approved):

PR #242, opened 2025-02-28
use sphinx internal linkcheck

linuxrider:
Checklist


.github/workflows/doc.yml
```diff
@@ -34,7 +34,5 @@ jobs:
       - name: detect broken links
         shell: pixi run -e dev bash -e {0}
         run: |
-          python -m http.server 8880 --directory doc/generated/html &
-          sleep 1
-          # repology.org only passes the check locally but not in the Github CI. see #169
-          linkchecker --check-extern http://localhost:8880/svgdigitizer/ --ignore-url="https://repology.org/*" --no-warnings
+          cd doc
+          make linkcheck
```
DunklesArchipel, line 38: Should we include this command in `pixi run doc` to perform the link checking also locally?
Or a separate one `pixi run linkcheck`

DunklesArchipel (approved):

.github/workflows/doc.yml
linuxrider: In ec-tools I have created a new pixi task for it. Is there a reason to not linkcheck when building the docs? Maybe request limits on the urls tested?

DunklesArchipel: We could construct something such as `doc = { depends-on = ["doc-build", "doc-lint"]  }, which allows you to execute tasks simultaneously or separately. You probably do not always want to validate the links to save time.

linuxrider: I think `make linkcheck` builds the docs and then checks the links.
So I added a separate pixi task for now.
In the CI we are build two times.

PR #253, opened 2025-07-09
Arm64

linuxrider:
Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ svgdigitizer = ["assets/template.svg"]
 
 [tool.pixi.project]
 channels = ["conda-forge"]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
     "pymupdf>=1.25,<2",
     "pyyaml>=6,<7",
     "scipy>=1.7,<2",
-    "svg.path>=4.1,<7",
+    "svg.path>=4.1,<8",
```
DunklesArchipel, line 24: This will require a news, such that I do not miss this change in the recipe in conda forge.

DunklesArchipel (approved):
Could you add news, also including the increase of the upper version limit of svg.path?

PR #262, opened 2025-10-19
Change --bibliography from type bolean to path

DunklesArchipel:
Checklist

The --bibliography now points specifically to a bib file and a key must be provided with --citation-key. For our purpose of converting svg to echemdb unitpackages, we retrieve the key directly from yaml metadata


svgdigitizer/entrypoint.py
```diff
@@ -157,35 +164,44 @@ def _create_svgplot(svg, sampling_interval, skewed):
     )
 
 
-def _create_bibliography(svg, metadata):
+def _create_bibliography(bibliography, citation_key, metadata):
     r"""
     Return a bibtex string built from a BIB file and a key provided in `metadata['source']['citationKey']`,
-    when both requirements are met. Otherwise an empty string is returned.
+    or from the provided `citation_key`when both requirements are met.
```
linuxrider:
```suggestion
    or from the provided `citation_key` when both requirements are met.
```

linuxrider (approved):

PR #267, opened 2025-11-15
fix paginate

linuxrider:
Checklist


DunklesArchipel (commented):
looks good. We should, however, at some point include this in the documentation.

Maybe also add an issue regarding nix.

PR #286, opened 2026-04-21
axis orientation with scalebars

DunklesArchipel:
Checklist

In some cases the orientation of a vertical scalebar is detected as x-axis. This has consequences for the CV module, which expects E or U  on the x-axis and I or j on the y-axis.

In principle the orientation does not matter at all, except for the code in CV. I think for now a more robust selection of the orientation (standard 2D plots, where the axis are perpendicular one to another) is sufficient. Otherwise one has to set x and y specifically in the CV module, which then means that we have to rewrite a number of properties from SVGFigure.


svgdigitizer/svg.py
```diff
@@ -507,12 +513,48 @@ def far(self):
...
+
+        rotation = self.label.rotation / 360 * 2 * pi
+        dir_x, dir_y = cos(rotation), sin(rotation)
+
+        def distance(p):
+            """The visual distance to the text is not strait forward to calculate.
+            We take the perpendicular distance to a line defined by anchor point an text direction.
+            Note: This might not cover all issues leading to secondary errors such as wrong axis
+            assignment.
+            """
```
DunklesArchipel:
```suggestion
            ""The distance of a line that is perpendicular to a line passing through an anchor point of a text object with a certain direction.
            Note: This might not cover all issues leading to secondary errors such as wrong axis
            assignment.
            """
```

DunklesArchipel: I am still not sure if this perfectly describes here. The function returns a distance but also a `parallel` value. The note is unclear here, especially what is meant by other errors and issues. Maybe it would be better to write which aspect is actually fixed by this approach. Maybe something such as `These values are required to calcukate...  and... mitigate an issue related to ....`

linuxrider (approved):

PR #296, opened 2026-06-24
Add create plot flag

### LINKED ISSUE #280: Flie not found error in paginate documentation
There is an error in the examples for the paginate method of the CLI in the documentation.

https://echemdb.github.io/svgdigitizer/workflow.html

DunklesArchipel:
The former implementation of creating a plot from the CLI made a plot appear as a popup. This does not seem to work on all OS.
Also the method was unspecific to the digitization approach used.

Thus it would make more sense to have a shared method, that can be evoked with a `--plot` flag.

The plot is shown with the "original" aspect ratio, which should simplify comparing the difgitized curves, to those found in the PDFs (such as in https://github.com/echemdb/electrochemistry-data)

Checklist

Fixes #280


doc/files/mustermann_2021_svgdigitizer_1/.gitignore
```diff
@@ -1,2 +1,18 @@
...
+
+# paginate outputs (the PDF has 4 pages -> _p1 .. _p4)
+mustermann_2021_svgdigitizer_1_p1.png
+mustermann_2021_svgdigitizer_1_p1.svg
+mustermann_2021_svgdigitizer_1_p2.png
+mustermann_2021_svgdigitizer_1_p2.svg
+mustermann_2021_svgdigitizer_1_p3.png
+mustermann_2021_svgdigitizer_1_p3.svg
+mustermann_2021_svgdigitizer_1_p4.png
+mustermann_2021_svgdigitizer_1_p4.svg
```
linuxrider, line 14: Do we need all pages or only the one with the plot?

DunklesArchipel: we do not need these per se, but they are generated from a CLI command, executed during the documentation build.

PR #301, opened 2026-07-17
Fix build_identifier crashing on UTF-8 characters outside latin-1

### LINKED ISSUE #300: Pdf.build_identifier crashes on UTF-8 characters outside latin-1 (e.g. ć, ž, —)
## Description

Since 0.14.5, `svgdigitizer.pdf.Pdf.build_identifier` translates LaTeX escape sequences to unicode via a latin-1 round-trip:

```python
entry.persons["author"][0].last_names[0].encode("latin-1").decode("latex+latin")
```

(and the same for the title field). This raises `UnicodeEncodeError` whenever the BibTeX entry already contains UTF-8 characters that are not representable in latin-1, e.g. `ć` (U+0107), `ž` (U+017E), or an em dash `—` (U+2014) in the title:

```python
>>> from pybtex.database import parse_string
>>> from svgdigitizer.pdf import Pdf
>>> bib = parse_string(
...     '@article{jovic_1996_test_1, author = {Jović, BM}, title = {Some title},'
...     ' year = {1996}, pages = {1--10}, journal = {J}}',
...     bib_format="bibtex",
... )
>>> Pdf.build_identifier(bib)
  ...
UnicodeEncodeError: 'latin-1' codec can't encode character 'ć' in position 4: ordinal not in range(256)
```

This breaks `validate-bib-keys` in [echemdb/electrochemistry-data](https://github.com/echemdb/electrochemistry-data), where the bibliography deliberately stores authors such as `Jović`, `Marinković`, and `Adžić` as plain UTF-8 (LaTeX accent encodings are explicitly forbidden there by `validate-bib-utf8`).

## Suggested fix

Only apply the latin-1/`latex+latin` round-trip when it is possible, and fall back to the string as-is (it is already unicode) otherwise, e.g.:

```python
def _latex_to_unicode(value):
    try:
        return value.encode("latin-1").decode("latex+latin")
    except UnicodeEncodeError:
        return value  # already plain unicode with non-latin-1 characters
```

Alternatively, encode with `errors="backslashreplace"`-style handling via latexcodec's own `latex+latin1` codec.

Observed with svgdigitizer 0.14.5.

DunklesArchipel:
The latin-1/latex+latin round-trip in Pdf.build_identifier raised a UnicodeEncodeError when the bibtex entry already contained plain UTF-8 characters not representable in latin-1 (e.g. ć, ž, or an em dash). Such values are now returned as-is since they are already unicode.

Fixes #300

Checklist

### DIFF
--- a/svgdigitizer/pdf.py
+++ b/svgdigitizer/pdf.py
@@ -265,6 +265,31 @@ class Pdf:
         self._rename(identifier + ".pdf")
 
     @staticmethod
+    def _latex_to_unicode(value):
+        r"""
+        Return `value` with LaTeX escape sequences translated to unicode.
+
+        Values that already contain unicode characters outside latin-1
+        are returned unchanged.
+
+        Examples::
+
+            >>> from svgdigitizer.pdf import Pdf
+            >>> Pdf._latex_to_unicode(r"\'Alvaro-Monta{\~n}a")
+            'Álvaro-Monta{ñ}a'
+            >>> Pdf._latex_to_unicode("Jović")
+            'Jović'
+        """
+        import latexcodec as _  # noqa: F401  # registers "latex+latin" codec
+
+        try:
+            return value.encode("latin-1").decode("latex+latin")
+        except UnicodeEncodeError:
+            # The value contains characters outside latin-1,
+            # so it is already plain unicode rather than LaTeX-escaped.
+            return value
+
+    @staticmethod
     def build_identifier(citation, skip_words=default_skip_words):
         """
         Build the entry identifier based on a bibtex citation provided as `BibliographyData` (pybtex).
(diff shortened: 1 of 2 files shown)


svgdigitizer/pdf.py
```diff
@@ -264,6 +264,31 @@ def rename_by_key(self):
...
+        """
+        import latexcodec as _  # noqa: F401  # registers "latex+latin" codec
+
+        try:
+            return value.encode("latin-1").decode("latex+latin")
+        except UnicodeEncodeError:
+            # The value contains characters outside latin-1,
+            # so it is already plain unicode rather than LaTeX-escaped.
+            return value
+
```
linuxrider, line 291:
```suggestion

```

svgdigitizer/pdf.py
```diff
@@ -353,23 +378,28 @@ def build_identifier(citation, skip_words=default_skip_words):
...
-        import latexcodec as _  # noqa: F401  # registers "latex+latin" codec
         from slugify import slugify
 
         entry = list(citation.entries.values())[0]
         first_author = (
-            entry.persons["author"][0]
-            .last_names[0]
-            .encode("latin-1")
-            .decode("latex+latin")
+            Pdf._latex_to_unicode(entry.persons["author"][0].last_names[0])
```
linuxrider, line 397:
```suggestion
            entry.persons["author"][0]
            .last_names[0]
            .encode("utf-8")
            .decode("latex+utf8")
```

svgdigitizer/pdf.py
```diff
@@ -353,23 +378,28 @@ def build_identifier(citation, skip_words=default_skip_words):
...
             .replace("{", "")
             .replace("}", "")
         )
         title_words = (
-            entry.fields["title"]
-            .encode("latin-1")
-            .decode("latex+latin")
+            Pdf._latex_to_unicode(entry.fields["title"])
```
linuxrider, line 402:
```suggestion
            entry.fields["title"]
            .encode("utf-8")
            .decode("latex+utf8")
```

linuxrider (changes requested):
I think we can get away with a trivial change.

PR #305, opened 2026-09-07
Add CITATION.cff

goelakash:
Adds a `CITATION.cff` so GitHub shows a "Cite this repository" button and citation tooling can read machine-readable metadata.

Author names and ORCIDs come from `paper.md`, and the licence from the LICENSE file. Validated with `cffconvert --validate` against schema 1.2.0.

Worth checking the given-name/family-name split per author, since some papers give a single `name` field.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,19 @@
+cff-version: 1.2.0
+message: If you use this software, please cite it as below.
+title: 'svgdigitizer: A Python package for recovering data from plots of scientific publications'
+type: software
+authors:
+- given-names: Johannes M.
+  family-names: Hermann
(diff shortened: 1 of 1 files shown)


goelakash:
Closing this. I opened it without asking first, and a citation file probably isn't something you need from a stranger. Sorry for the noise.

DunklesArchipel:
Thanks for reaching out. I was not aware that such a citation file exists. One problem I see is with the DOI that can be provided. Once a new repository version is created zendo will make a new release. The zenodo DOI has to be updated after the new zenodo release. 
On the other hand the zenodo DOI with a link is already provided in the readme. On that zenodo page the citation can be exported in various formats. Probably that is the best place to avoid circular dependencies.

